PR #1, opened 2020-03-09
Initialize Portunus using Zygoat

Ian-MacLeod:
And we're off!

The flake8 pre-commit hook was causing problems so I disabled it for now. Will re-enable when https://github.com/bequest/zygoat/pull/21 gets merged. `venv/` was not added in the `.gitignore` by default, so I added that, too. Will make a PR for that in Zygoat shortly.

Encountered some warnings during dependency installation, mostly related to `wui` dependencies:

Screenshot from 2020-03-09 16-33-33

Other than that, everything seems good to go!


frontend/.eslintrc.js
```diff
@@ -0,0 +1,68 @@
...
+        when: "never",
+        children: true,
+        allowMultiline: true
+      }
+    ]
+  },
+
+  "import/resolver": {
+    "babel-module": {
+      root: ["./src"],
```
markrawls: There is no `src` directory in a `NextJS` app (although it does support that). Zygoat doesn't generate that by default.

frontend/.eslintrc.js
```diff
@@ -0,0 +1,68 @@
...
+      alias: {
+        "@wui": "@bequestinc/wui"
```
markrawls: In addition to that, we're using `@@` as the alias for the source root

backend/backend/settings/zygoat_settings.py
```diff
@@ -13,92 +14,86 @@
...
-    {
-        'NAME': 'django.contrib.auth.password_validation.CommonPasswordValidator',
-    },
-    {
-        'NAME': 'django.contrib.auth.password_validation.NumericPasswordValidator',
-    },
+    {"NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",},
+    {"NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",},
+    {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator",},
+    {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator",},
```
Ian-MacLeod: Hmm, not sure how I feel about what Black is doing here with the trailing commas. I suppose it is being extra careful not to make any meaningful changes. Will this even pass flake8?

markrawls: Generally speaking I've found it best not to add trailing commas and just let `Black` handle it on save, same deal with `prettier`.

frontend/.eslintrc.js
swang192: Oh those were my changes. Should I just get rid of `import/resolver` altogether?

backend/backend/settings/zygoat_settings.py
```diff
@@ -0,0 +1,138 @@
...
+
+SESSION_COOKIE_DOMAIN = SHARED_DOMAIN
+
+SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"
+
+SESSION_COOKIE_AGE = 604800
+
+if not DEBUG:
+    SESSION_COOKIE_SECURE = True
+    CSRF_COOKIE_SECURE = True
```
markrawls: Since we have a decoupled frontend that is no longer being served by the backend, the backend itself cannot directly set cookies on requests. How do you plan to make these work?

Ian-MacLeod:
@markrawls I'll make those updates here manually for now, and a PR on Zygoat for the future, too.

markrawls:
> Encountered some warnings during dependency installation, mostly related to wui dependencies:

Who wants to be the brave soul that makes the PR to run upgrades on that?

frontend/.eslintrc.js
markrawls: I would say probably add the `@@` alias to that section and change the root directive to just `./` and test it out

swang192:
> > Encountered some warnings during dependency installation, mostly related to wui dependencies:
> 
> Who wants to be the brave soul that makes the PR to run upgrades on that?

I'll look

backend/backend/settings/zygoat_settings.py
Ian-MacLeod: Well, one option is to pass the cookies along when we are proxying the requests/repsonses. Another option is to not use cookies at all for our typical API calls, although I think we will still want to use them on Portunus specifically for SSO.

Ian-MacLeod:
Re-initialized with the latest version of Zygoat

markrawls:
Let's get the latest version out and then do it OOOOOOONE more time

Ian-MacLeod:
Re-initialized using Zygoat 0.1.4

kborer:
lgtm

Issue #2, opened 2020-03-11
Create frontend register/login/password reset pages


Ian-MacLeod:
I'm still figuring out SSO stuff, but we can start building the frontend auth pages in the meantime. From what I can tell looking at the designs, they can look exactly like the ones we have for Willing, but with the logo swapped out. CSRF token should be passed like we currently do on Willing. If a `next` url parameter is passed, the frontend should also pass that in the request and the user should be redirected to it's value after a successful login.

PR #11, opened 2020-03-25
Add login and register pages

swang192:
Closes part of #2. Includes recent zygoat changes (0.2.0 to 0.2.2)


frontend/components/Layout.js
```diff
@@ -0,0 +1,28 @@
...
+        <Panel>
+          {children}
+        </Panel>
+        <Spacer v={92} />
+      </Grid>
+    </div>
+  );
+}
+
+export default Layout;
```
cviveiro: missing newline at end of file

frontend/components/Register.js
```diff
@@ -0,0 +1,95 @@
...
+import Link from 'next/link';
+import { register } from '../utils/API';
+
+import Form from '@wui/layout/form';
+import Button from '@wui/input/button';
+import Spacer from '@wui/layout/spacer';
+import Textbox from '@wui/input/textbox';
+import Typography from '@wui/basics/typography';
+
+// const registerErrors = {
```
cviveiro: commented out code that should probably be removed or used?

frontend/components/Layout.js
cviveiro: this is in more of the files too, you should make sure your editor of choice is configured to add the newline at EOF

Ian-MacLeod:
Shouldn't the pre-commit hooks be catching the EOF newlines?

cviveiro:
Yeah, but you do have to install them yourself (it's not automatic) and I'm guessing they weren't installed. `pre-commit install` will get them running so they catch this.

frontend/components/Register.js
swang192: These are errors from willing app that should be in portunus too, but I guess we can add them later too.

swang192:
Newlines fixed

backend/authentication/views.py
```diff
@@ -10,17 +12,19 @@
 
 @require_POST
 def register(request):
+    data = json.loads(request.body)
     user = User.objects.create_user(
-        email=request.POST.get("email"), password=request.POST.get("password"),
+        email=data['email'], password=data['password'],
```
Ian-MacLeod: It looks like Black hasn't been run on this (or it would convert these to double quotes). Try doing `pre-commit install` and then `pre-commit run --all-files`

frontend/utils/API.js
```diff
@@ -0,0 +1,50 @@
...
+      config.data = params;
+    } else {
+      config.headers['Content-Type'] = 'application/json';
+    }
+  }
+
+  return config;
+});
+
+const register = async (email, password) => {
```
Ian-MacLeod: This approach matches what we do on Willing and seems fine. What do we think about passing a single argument for `payload` or `data`? Doesn't make a big difference here, but when there are more fields to submit I think that using separate arguments becomes a bit cumbersome.

frontend/utils/API.js
```diff
@@ -0,0 +1,50 @@
...
+  if (config.data) {
+    // Convert objects to url search params, since that's what we're posting to the backend.
+    // Strings are assumed to be JSON.
+    if (typeof config.data === 'object') {
+      const params = new URLSearchParams();
+      for (const [key, value] of Object.entries(config.data)) {
+        params.append(key, value);
+      }
+      config.data = params;
+    } else {
```
Ian-MacLeod: Since the proxy server is converting this to JSON anyways, we might want to get rid of this and instead call JSON.stringify on the object and submit everything as `application/json`.

frontend/utils/API.js
```diff
@@ -0,0 +1,50 @@
...
+  return config;
+});
+
+const register = async (email, password) => {
+  return await API.post('auth/register', { email, password });
+}
+
+const login = async (email, password) => {
+  return await API.post('auth/login', { email, password });
+}
```
Ian-MacLeod: This again matches how we do things in the Willing app, but I think that we don't need to use async/await here. Just returning the promise from `API.post` directly seems simpler.

Ian-MacLeod:
Minor point for the future that I don't think is necessary to change in this PR: I think it would be better for reviewing and for having a meaningful git history if we put things like upgrading the zygoat version in their own commit.

frontend/utils/API.js
swang192: Agreed, changed to just pass in a single payload. Guess the only difference is where we can go to look at what params are passed into the request (just the component now vs component or API.js before).

swang192:
Addressed comments and cleaned up into separate commits

frontend/utils/API.js
```diff
@@ -0,0 +1,29 @@
...
+
+const API = axios.create({
+  baseURL: '/api/',
+  responseType: 'json',
+  headers: { 'X-REQUESTED-WITH': 'XMLHttpRequest' },
+})
+
+API.interceptors.request.use(config => {
+  config.headers['X-CSRFToken'] = Cookies.get('csrftoken');
+  config.headers['Content-Type'] = 'application/json';
```
Ian-MacLeod: Nit: Since this is static now, it can probably go in the axios initialization above.

swang192: Yeah good point. Updated

PR #13, opened 2020-03-25
Add in change/reset password views

### LINKED ISSUE #3: Create backend auth/password reset views
These will likely have to be modified once we have settled on our SSO mechanism, but I think it would still be productive to get them up and going in the meantime.

cviveiro:
Closes #3

Tested all the views I could as much as I could in not production. The password reset stuff was hard because auth doesn't look to be working when `PRODUCTION == True` so I couldn't send myself the email. I did see the "email" though when in debug mode, and could get to the complete page and reset my password there.


backend/authentication/utils.py
```diff
@@ -12,3 +36,52 @@ def login_user(request, user):
...
     User.objects.create_user(portunus_uuid=portunus_uuid)
+
+
+def check_and_change_password(drf_request, user, new_password):
+    if not is_password_valid(new_password):
+        return make_response(False, error_msg="invalid_password")
+
+    user.set_password(new_password)
+    user.save()
+    update_session_auth_hash(drf_request, user)
```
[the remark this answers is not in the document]
Ian-MacLeod: On second thought, this won't work - if we want to end existing sessions when the password is changed we should use the [blacklist app](https://django-rest-framework-simplejwt.readthedocs.io/en/latest/blacklist_app.html) that drf-simplejwt provides.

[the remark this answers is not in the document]
cviveiro: I think the blacklist will also fix logging out (I was able to logout but still hit `is_authenticated` because the token was still good), it was one of the solutions mentioned to fix that problem. So I'll add that in.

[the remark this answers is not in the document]
Ian-MacLeod: Clearing the refresh token from the cookies should do the trick, but I think it does makes sense to blacklist the token on logout anyways.

backend/authentication/utils.py
```diff
@@ -1,14 +1,109 @@
...
+
+
+def _send_email(to_list, subject, template_name, data):
+    email_txt = render_to_string(f"reset_password.txt", data)
+    email_html = render_to_string(f"reset_password.html", data)
+    msg = EmailMultiAlternatives(
+        subject=subject,
+        body=email_txt,
+        # TODO change
+        from_email="Willing <hello@willing.com>",
```
cviveiro: Oh, do we have an email we should change this to?

kborer: change it to a shared constant, but the leave the value the same until we have control of the `legalplans.com` domain next month

cviveiro:
Ok, changes made.

We now blacklist all of the user's tokens when they change their password. Tested this, works as expected.

We also use the existing blacklist framework to create "one time" tokens that are registered to specific users and use those tokens to reset passwords. Seemed best to use a framework we were already using, rather than build out support for a new one. Plus they're tied to users automatically, so I think that could give us some extra security. Tested all the ways that tokens can be invalid, and then tested to make sure that it works and you can reset your password and that logins work with the new password.

Note: we will need to add in a scheduled job to flush expired tokens (`./manage.py flushexpiredtokens`) that's provided with the blacklist app. It's suggested to run it daily. I'll create a GH issue for it once we verify that we definitely want to move in this direction.

backend/authentication/utils.py
```diff
@@ -14,5 +42,104 @@ def login_user(request, user):
...
+    # Make sure that we created and sent this token and that it matches a record
+    # we have in our db.
+    try:
+        saved_token = OutstandingToken.objects.filter(
+            token=token_str, user__portunus_uuid=user.portunus_uuid
+        )[0]
+    except IndexError:
+        # No matches found, either the user id is a mismatch or the token string
+        # doesn't match.
+        print("No matching existing token found.")
```
Ian-MacLeod: rm

backend/authentication/utils.py
```diff
@@ -14,5 +42,104 @@ def login_user(request, user):
...
+        return False
+
+    if saved_token.created_at < make_aware(
+        datetime.now() - timedelta(minutes=RESET_LINK_VALID_MINUTES)
+    ):
+        print("Token has expired.")
```
Ian-MacLeod: rm

backend/authentication/utils.py
```diff
@@ -14,5 +42,104 @@ def login_user(request, user):
...
+        return False
+
+    try:
+        token = RefreshToken(token_str)
+    except TokenError:
+        print("Token has been blacklisted.")
```
Ian-MacLeod: rm

backend/authentication/utils.py
```diff
@@ -14,5 +42,104 @@ def login_user(request, user):
...
+
+def send_password_reset(user):
+    # Make a refresh token that will serve as a one-time token linked to this user.
+    # NOTE: we can't set the exp here since that changes the token string without
+    # changing the underlying value in the database, which we need to check against
+    # to verify some other stuff. We'll check by `created_at` when the link is clicked.
+    token = RefreshToken.for_user(user)
+    reset_url = parse.urljoin(
+        settings.BASE_URL, f"/api/auth/reset-password/complete/{user.portunus_uuid}/{token}/",
+    )
```
Ian-MacLeod: Does `urljoin` handle escaping, too?

backend/authentication/views.py
```diff
@@ -1,36 +1,99 @@
...
     login_user(request, user)
-    return HttpResponse("success", status=status.HTTP_200_OK)
+    return make_response(True)
 
 
 @require_POST
 def logout(request):
+    # First, get the session refresh token (before it gets cleared out on logout) and
+    # blacklist it.
+    blacklist_token(request.session.get(REFRESH_TOKEN_SESSION_KEY))
```
Ian-MacLeod: I think it would be good to blacklist all refresh tokens for the user here.

backend/authentication/utils.py
```diff
@@ -14,5 +42,104 @@ def login_user(request, user):
...
+    token.blacklist()
+    return True
+
+
+def send_password_reset(user):
+    # Make a refresh token that will serve as a one-time token linked to this user.
+    # NOTE: we can't set the exp here since that changes the token string without
+    # changing the underlying value in the database, which we need to check against
+    # to verify some other stuff. We'll check by `created_at` when the link is clicked.
+    token = RefreshToken.for_user(user)
```
Ian-MacLeod: This might be kind of nitty, but it seems a bit awkward to re-use the `RefreshToken` class for this. Can we instead create and use a `BlacklistableToken` class that inherits from both `BlacklistMixin` and `Token`.

backend/authentication/utils.py
cviveiro: Question: I left these in because it would likely be very helpful if we need to debug to see where we go wrong when checking a token. Happy to take these out, but I'm not sure whether the help for debugging would be necessary to add in some other way. Is there a place for debugging info like this? Or do we have other things in place to make it so that we can test through and figure out what happens?

Ian-MacLeod: If we want to keep debugging statements like this I think using Django's [logging](https://docs.djangoproject.com/en/3.0/topics/logging/) system would probably be best.

cviveiro:
Changes made and tested.

I did have a problem with the changes that used `request.body`. That seemed super fragile as it was complaining about not being able to access that value twice and something in the middleware was using it, so I switched it back to just using the POST data.

backend/authentication/utils.py
```diff
@@ -23,6 +23,8 @@
 REFRESH_TOKEN_SESSION_KEY = "refresh_token"
 
 RESET_LINK_VALID_MINUTES = 5
+# TODO change this when we have control of the `legalplans.com` domain.
+RESET_LINK_FROM_EMAIL = "Willing <hello@willing.com>"
```
Ian-MacLeod: Nit: I think something more general is preferable here: just `FROM_EMAIL` maybe

cviveiro:
Changes made and tested.

PR #15, opened 2020-03-26
Cycle csrf

Ian-MacLeod:


cviveiro:
LGTM, thanks for doing, saw that note about rotation too.

Issue #12, opened 2020-03-25
Acquire CSRF token on page load


Ian-MacLeod:
Since the frontend is not served by the backend, we need to use an api endpoint to set the cookie after the initial page load.

PR #16, opened 2020-03-29
Add endpoint to set csrf cookie

cviveiro:
Adds backend endpoint for #12.

Tested, works as expected.

### DIFF
--- a/backend/backend/views.py
+++ b/backend/backend/views.py
@@ -1,7 +1,6 @@
-from django.middleware.csrf import get_token
 from django.http import HttpResponse
-from django.views.decorators.http import require_POST
-from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_http_methods
+from django.views.decorators.csrf import ensure_csrf_cookie
 from rest_framework.decorators import api_view, permission_classes
 from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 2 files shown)


backend/backend/views.py
```diff
@@ -1,10 +1,23 @@
...
 def is_authenticated(request):
     content = {"email": request.user.email}
     return Response(content)
+
+
+@csrf_exempt
+@require_POST
+def set_csrf(request):
+    # This will set the csrf cookie data as a side effect.
+    get_token(request)
```
Ian-MacLeod: It looks like there is a built in decorator that will handle this: [ensure_csrf_cookie](https://docs.djangoproject.com/en/3.0/ref/csrf/#django.views.decorators.csrf.ensure_csrf_cookie)

backend/backend/views.py
```diff
@@ -1,10 +1,23 @@
...
 
 @api_view(["GET"])
 @permission_classes([IsAuthenticated])
 def is_authenticated(request):
     content = {"email": request.user.email}
     return Response(content)
+
+
+@csrf_exempt
+@require_POST
```
Ian-MacLeod: I think GET would be OK here.

cviveiro:
Changes made and tested.

PR #20, opened 2020-04-01
BUILD: Add bandit and upgrade flake8-quotes

ambikads:
CI in codebuild was failing due to missing bandit. Turns out this https://github.com/bequest/zygoat/pull/47 change from zygoat was needed.
The `flake8-quotes` got upgraded when I ran zygoat.


cviveiro:
LGTM

PR #21, opened 2020-04-01
Set CSRF cookie on page load

### LINKED ISSUE #12: Acquire CSRF token on page load

cviveiro:
Closes #12


backend/backend/views.py
```diff
@@ -14,6 +14,7 @@ def is_authenticated(request):
     return Response(content)
 
 
+@csrf_exempt
 @ensure_csrf_cookie
 @require_http_methods(["GET", "POST"])
```
Ian-MacLeod, line 19: Nit, if we disallow POST on this endpoint we don't have to mark it as csrf_exempt. I think that's a bit nicer.

PR #22, opened 2020-04-02
Handle validation errors on login/reset views

Ian-MacLeod:
Since we are using some built in Django validators and they provide field-level errors with user-friendly strings I thought I would try leveraging that for our 400 responses. Seems like a simple approach for form-based views. Let me know what you think.

### DIFF
--- a/backend/authentication/serializers.py
+++ b/backend/authentication/serializers.py
@@ -1,0 +1,40 @@
+from django.contrib.auth import authenticate
+from django.utils.translation import gettext as _
+from django.contrib.auth.password_validation import validate_password
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+
+from .models import User
+
+
+class RegistrationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = ("email", "password")
+        extra_kwargs = {"password": {"write_only": True}}
+
+    def validate_password(self, value):
+        validate_password(value)
+        return value
+
+    def create(self, validated_data):
+        return User.objects.create_user(
+            email=validated_data["email"], password=validated_data["password"],
+        )
+
+
+class LoginSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = ("email", "password")
+        extra_kwargs = {"password": {"write_only": True}, "email": {"validators": []}}
+
+    def validate(self, data):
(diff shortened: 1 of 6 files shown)


backend/authentication/utils.py
```diff
@@ -25,12 +27,16 @@
...
 
-def make_response(success, error_msg=None, status=None):
+def make_response(success, data=None, status=None):
     if not status:
         status = status_codes.HTTP_200_OK if success else status_codes.HTTP_400_BAD_REQUEST
 
-    msg = "success" if success else error_msg or "failure"
-    return HttpResponse(msg, status=status)
+    if not data:
+        data = {"success": success}
```
cviveiro, line 35: Why don't we just always update data with this field? This would make it so that you have to add it yourself if you want it but also have other data, when it seems like it's a general thing that would be good to add in for all responses.

backend/authentication/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+
+class LoginSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = ("email", "password")
+        extra_kwargs = {"password": {"write_only": True}, "email": {"validators": []}}
+
+    def __init__(self, *args, **kwargs):
+        self.user = None
+        print(args, kwargs)
```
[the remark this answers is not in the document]
Ian-MacLeod: Whoops, It's not important - I had intended to remove this. Updated.

frontend/utils/hooks.js
```diff
@@ -1,9 +1,8 @@
 import { useState } from 'react';
 
-const useInputFieldState = initial => {
+// eslint-disable-next-line import/prefer-default-export
```
cviveiro, line 3: Why make this change if it means the linter gets upset? I get making stylistic changes but not really when that means you have to carve out an exception to the linting rules.

Ian-MacLeod: I guess we could leave it as a default export for now and change to named exports when we add another hook to the file. The reason I changed this is that it seems unnatural to me for a file intended to define hooks to export a single hook as a default export. I don't feel too strongly about it, though.

Ian-MacLeod: Made it a default export for now.

cviveiro:
LGTM!

PR #23, opened 2020-04-02
Redirect on successful authentication

Ian-MacLeod:
Only the last commit is part of this PR. To be merged after #22. Passing the redirect URL through the backend is a little awkward, but I thought it would be good to make sure we don't redirect to arbitrary URLs.

### DIFF
--- a/backend/authentication/utils.py
+++ b/backend/authentication/utils.py
@@ -1,8 +1,10 @@
 import json
 from urllib import parse
+from urllib.parse import urlparse
 
 from anymail.exceptions import AnymailError
 from django.conf import settings
+from django.core.validators import URLValidator
 from django.http import HttpResponse
 from django.core.mail import EmailMultiAlternatives
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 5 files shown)


backend/authentication/utils.py
```diff
@@ -46,6 +48,17 @@ def login_user(request, user):
...
 
+def get_valid_redirect_url(url):
+    is_valid = URLValidator()
+    try:
+        is_valid(url)
+        if urlparse(url).hostname not in settings.VALID_REDIRECT_HOSTNAMES:
+            raise ValidationError
+        return url
+    except ValidationError:
+        return settings.DEFAULT_REDIRECT_URL
```
cviveiro, line 59: It seems like we should either require a default prod for this or shouldn't redirect here, since it would be weird to send to localhost during prod? Seems like some sort of catch is needed for prod.

backend/backend/settings/__init__.py
```diff
@@ -1,5 +1,13 @@
 from .zygoat_settings import *  # noqa
 
+
+def prod_required_env(key, default):
```
cviveiro, line 4: This is awesome, and I wonder if it shouldn't be included in `zygoat_settings.py` instead so that it's available to all zygoat apps.

backend/authentication/utils.py
Ian-MacLeod: Yes, for sure. Updated.

backend/backend/settings/__init__.py
Ian-MacLeod: That seems reasonable to me, probably should be used on some of the default settings, too. I'll make a PR

frontend/components/Login.js
```diff
@@ -83,7 +91,7 @@ const Login = () => {
       </Typography>
       <Typography>
         Forgot your password?&nbsp;
-        <Link href="/reset-password">
```
cviveiro: Naive question: won't this link not work now?

Ian-MacLeod: Huh, as of master the reset password page doesn't exist so this was a dummy value anyways - I don't see any reason for this to touch it, though. Updated.

cviveiro:
Also LGTM!

PR #28, opened 2020-04-09
Ignore backend settings file when run from 'backend' dir with Flake8

cviveiro:
This is pretty hacky, but it'll get linting to run and pass correctly on the CI build. If this is an ok solution, I'll add it to the zygoat repo file. Right now, linting is failing on all the new PRs. I'd swear it didn't on the original PR adding in codebuild stuff.... ?? But now that commit is marked as failed as well.

The problem is that the path to the backend file is specified from the root dir, but the flake8 CI command is run from the `backend` dir. Tested this and it 1) didn't raise issues from non-existent files to ignore and 2) worked. 

I didn't find a better solution, but didn't look hard and also I don't think there's going to be a much better one on the flake8 side. But happy to close this if there's another solution that's better.

### DIFF
--- a/backend/.flake8
+++ b/backend/.flake8
@@ -1,6 +1,8 @@
 [flake8]
 max-line-length = 95
 exclude = .git, __pycache, .pycharm_helpers, */migrations/*, node_modules, bower_components, venv
-per-file-ignores = backend/backend/settings/__init__.py:F405
+per-file-ignores =
+    backend/backend/settings/__init__.py:F405
+    backend/settings/__init__.py:F405
 extend-ignore = E126,B003,Q001,A003,E203,E501,Q000,C812,E231,C819
 # See https://github.com/PyCQA/pycodestyle/issues/373


ambikads:
The original codebuild did not have flake8 tests. Looks like what you have done is the recommended approach to ignore specific checks on individual files. I am okay with this.

ambikads:
ah I get it now with the path

backend/.flake8
```diff
@@ -1,6 +1,8 @@
 [flake8]
 max-line-length = 95
 exclude = .git, __pycache, .pycharm_helpers, */migrations/*, node_modules, bower_components, venv
-per-file-ignores = backend/backend/settings/__init__.py:F405
+per-file-ignores =
+    backend/backend/settings/__init__.py:F405
```
ambikads, line 5: is this needed to get flake8 work from the top level? is that something we want?

cviveiro: Yes, for pre-commit hooks (which I hope more people than just me are using). Those are always run from the top level repo directory.

cviveiro:
i don't know if that's possible, I can't see anything suggesting it is. 

Note that we really shouldn't be adding any more files to this. I can't imagine another file where we'd want to completely ignore something linting. If we do, it's either something Zygoat manages in which case we can add once, or it really shouldn't be that way. In my opinion.

cviveiro:
We now have frontend linting failures that are failing :(

Ian-MacLeod:
I think #29 should fix the frontend linting errors.

Ian-MacLeod:
I'm going to go ahead and merge this as the frontend linting stuff is clearly unrelated.

PR #32, opened 2020-04-13
Add in better error messages

### LINKED ISSUE #30: Show better errors on the frontend on failures
https://github.com/bequest/portunus/pull/27#discussion_r406876663

cviveiro:
Closes #30 

This adds clear password validation errors on the reset page as well as some other places and fixes some backend issues that came about from a past PR. Tested all of the reset functionality.

### DIFF
--- a/backend/authentication/utils.py
+++ b/backend/authentication/utils.py
@@ -88,11 +88,7 @@ def check_and_change_password(drf_request, user, new_password):
         validate_password(new_password)
     except ValidationError as e:
         return make_response(
-            False,
-            {
-                "error": INVALID_PASSWORD,
-                "validation_error": str(e.error_list[0]).strip("[]'\""),
-            },
+            False, {"error": INVALID_PASSWORD, "validation_error": e.messages[0],},
         )
 
     blacklist_user_tokens(user)
--- a/backend/authentication/utils.py
+++ b/backend/authentication/utils.py
@@ -66,7 +66,8 @@ def get_valid_redirect_url(url):
 def blacklist_token(token_str):
     if token_str:
         try:
-            token = RefreshToken(token_str)
+            # We don't verify the token so that, no matter what, we blacklist it.
+            token = RefreshToken(token_str, verify=False)
             token.blacklist()
         except TokenError:
             # Token is already blacklisted.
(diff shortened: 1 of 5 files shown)


backend/authentication/utils.py
```diff
@@ -66,7 +66,8 @@ def get_valid_redirect_url(url):
 def blacklist_token(token_str):
     if token_str:
         try:
-            token = RefreshToken(token_str)
+            # We don't verify the token so that, no matter what, we blacklist it.
+            token = RefreshToken(token_str, verify=False)
```
Ian-MacLeod, line 70: We could also take this out of the try block if we want to make sure that we don't ignore errors for this line.

backend/authentication/utils.py
```diff
@@ -118,12 +119,10 @@ def check_onetime_token(token_str, user):
 
     try:
         # This will verify that the token is valid and hasn't expired.
-        token = ResetToken(token_str)
+        _ = ResetToken(token_str)
     except TokenError:
         return False
 
-    # Blacklist the token so that it can only get used once.
-    token.blacklist()
```
Ian-MacLeod, line 126: Do we not want to blacklist these after use anymore?

backend/authentication/utils.py
```diff
@@ -83,8 +84,16 @@ def create_user(portunus_uuid):
...
-    if not is_password_valid(new_password):
-        return make_response(False, error_msg=INVALID_PASSWORD)
+    try:
+        validate_password(new_password)
+    except ValidationError as e:
+        return make_response(
+            False,
+            {
+                "error": INVALID_PASSWORD,
+                "validation_error": str(e.error_list[0]).strip("[]'\""),
```
Ian-MacLeod, line 94: `str(e.error_list[0]).strip("[]'\"")` -> `e.message_list[0]`

frontend/components/CompleteResetPassword.js
```diff
@@ -48,7 +48,11 @@ const ResetPasswordComplete = ({ uuid, token }) => {
       setValidationError(null);
       setSuccess(true);
     } catch (error) {
-      setValidationError('ERROR');
+      const errorMsg =
+        error.response.data.error === 'invalid_password'
```
Ian-MacLeod, line 52: We should use a constant for the invalid password error code.

backend/authentication/utils.py
cviveiro: So I took this out so that we'd blacklist the token after we have validated the password and so that this just becomes a check function. We actually blacklist this token when we change the password and `blacklist_user_tokens`, so it gets taken care of then.

backend/authentication/utils.py
cviveiro: I tried that at first and got a json error because the data looked like this:
```
'validation_error': ValidationError(['This password is too short. It must contain at least 7 characters.'])'
```
Just reverified, that still causes an error.

Ian-MacLeod: Huh, really. Even using `message_list`, not `error_list`?

cviveiro:
```
backend_1   | AttributeError: 'ValidationError' object has no attribute 'message_list'
```
Looking at the [source code](https://docs.djangoproject.com/en/3.0/_modules/django/core/exceptions/#ValidationError) I didn't see anything better to use with just the raw `Error`?

Ian-MacLeod: Oh, sorry, it looks like it should be `e.messages[0]`. Does that work?

cviveiro: Don't know how I missed that >.< Thanks! Updated and tested.

Ian-MacLeod: No worries! Looks good

PR #33, opened 2020-04-14
Deployment configuration

markrawls:


backend/requirements.txt
```diff
@@ -8,3 +8,4 @@ djangorestframework-simplejwt==4.4.0
 cryptography==2.8
 simplejwt-extensions==0.1.0
 django-anymail==7.0.0
+zappa==0.51.0
```
cviveiro, line 11: Should this be standard and managed by zygoat?

backend/zappa_settings.json
```diff
@@ -0,0 +1,22 @@
+{
+  "base": {
+    "aws_region": "us-east-1",
+    "django_settings": "backend.settings",
+    "profile_name": "default",
+    "project_name": "portunus-backend",
+    "runtime": "python3.7",
+    "s3_bucket": "com.willing.portunus-deployments"
```
cviveiro, line 8: Missing comma

backend/requirements.txt
cviveiro: Oh no, that's right, we took deployment out. Nevermind.

markrawls: I am looking at adding this into our plugin repository however, but that's outside the scope of this PR.

backend/zappa_settings.json
markrawls: Ran prettier against the file to make it a little more consistent.

PR #34, opened 2020-04-14
Add social auth

Ian-MacLeod:
This adds Facebook and Google login. I also refactored the login/register components because they were sharing a bunch of logic. I used a development version of zygoat to update so we shouldn't merge until https://github.com/MetLifeLegalPlans/zygoat/pull/67 is merged, but this is ready for review.


backend/authentication/social_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    return provider == user.social_login_provider
+
+
+def _get_facebook_access_token():
+    return (
+        requests.get(
+            "https://graph.facebook.com/v3.0/oauth/access_token",
+            params={
+                "client_id": "275311629877038",
+                "client_secret": settings.FACEBOOK_OAUTH_APP_SECRET,
```
cviveiro, line 36: Think this should this also be a prod required setting like the google app id?

backend/authentication/models.py
```diff
@@ -8,9 +8,20 @@
 
 
 class User(AbstractUser):
+    PROVIDER_FACEBOOK = "facebook"
+    PROVIDER_GOOGLE = "google"
+
+    PROVIDER_CHOICES = (
+        (PROVIDER_FACEBOOK, "Facebook"),
```
cviveiro, line 15: Naive question: why allow for different caps policies on names? why not just standardize to all lower caps?

backend/authentication/serializers.py
```diff
@@ -38,3 +39,33 @@ def validate(self, data):
...
         return self.user
+
+
+class SocialAuthSerializer(serializers.Serializer):
+    provider = serializers.CharField(required=False)
+    token = serializers.CharField(required=False)
+    email = serializers.CharField(required=False)
+
+    def validate(self, data):
+        if not is_valid_token(data.get("email"), data.get("token"), data.get("provider")):
```
cviveiro, line 50: I don't know how this gets hit but you might consider making sure `provider` is all lower caps, so you never have a fallthrough when validating tokens if we get "Facebook" instead of "facebook"

frontend/components/SocialAuth/index.js
```diff
@@ -0,0 +1,88 @@
...
+import Spacer from '@wui/layout/spacer';
+
+import { socialAuth } from '@@/utils/API';
+import Facebook from './Facebook';
+import Google from './Google';
+
+const ERRORS = {
+  idpiframe_initialization_failed: 'Cookies must be enabled for you to login using this method.',
+  popup_closed_by_user: "Please try again and don't close any popup windows that may have opened.",
+  access_denied: 'Please grant the requested permissions the popup window.',
```
cviveiro, line 15: '...permissions in the popup window.'

backend/authentication/social_utils.py
Ian-MacLeod: Yeah, I think that's a good idea. Updated.

backend/authentication/models.py
Ian-MacLeod: The second values aren't alternatives to the first ones - they are user friendly representations used for the field in the Django admin. See: https://docs.djangoproject.com/en/3.0/ref/models/fields/#choices

backend/authentication/serializers.py
Ian-MacLeod: I suppose it wouldn't hurt to lowercase it. Updated

backend/authentication/models.py
cviveiro: Ahhhh ok! TIL

cviveiro:
LGTM but I am not qualified to really review the frontend bits.

frontend/assets/stylesheets/averta.css
```diff
@@ -0,0 +1,58 @@
+/**
```
markrawls, line 1: There's a different way that Averta has to be dropped into our NextJS apps, it's part of one of my PRs right now. Take this out for now and I'll get it in.

frontend/components/AuthBase.js
```diff
@@ -0,0 +1,144 @@
...
+    if (error.response && error.response.data) {
+      setInputErrors(error.response.data);
+    } else {
+      setInputErrors({ non_field_errors: 'An unknown error has occurred. Please try again.' });
+    }
+    setProcessing(false);
+  };
+
+  const handleSubmit = async e => {
+    e.preventDefault();
```
markrawls, line 55: `preventDefault()` can potentially be a race condition in async functions

frontend/components/AuthBase.js
```diff
@@ -0,0 +1,144 @@
...
+  const handleError = error => {
+    if (error.response && error.response.data) {
+      setInputErrors(error.response.data);
+    } else {
+      setInputErrors({ non_field_errors: 'An unknown error has occurred. Please try again.' });
+    }
+    setProcessing(false);
+  };
+
+  const handleSubmit = async e => {
```
markrawls, line 54: To avoid this race condition, just remove `async` from the function definition. You're not using `await` anywhere, and `Promise.then` is perfectly usable in synchronous functions.

frontend/assets/stylesheets/averta.css
markrawls: Ah brain fart - this is Portunus. Take it out anyways and I'll open a PR.

frontend/utils/API.js
```diff
@@ -21,6 +21,8 @@ const register = payload => API.post('auth/register/', payload);
 
 const login = payload => API.post('auth/login/', payload);
 
+const socialAuth = payload => API.post('auth/social-auth/', payload);
+
```
markrawls, line 25: Not a CR, but a stateless functional API with predefined actions is just beautiful

frontend/components/SocialAuth/index.js
```diff
@@ -0,0 +1,88 @@
...
+        }
+        render={Google({ disabled: processing })}
+      />
+
+      <Spacer v={14} />
+
+      <FacebookLogin
+        fields="email"
+        autoLoad={false}
+        appId="275311629877038"
```
markrawls, line 71: Make a `@@/constants` file please

frontend/components/SocialAuth/index.js
```diff
@@ -0,0 +1,88 @@
...
+      .catch(handleError);
+  };
+
+  return (
+    <>
+      <GoogleLogin
+        buttonText="Google"
+        scope="openid email"
+        onFailure={onFailure}
+        clientId="685797762223-edandm45c2023v4bvcqanggdkhgb3ims.apps.googleusercontent.com"
```
markrawls, line 53: `@@/constants` here as well please

frontend/components/AuthBase.js
Ian-MacLeod: Good point, I neglected to remove the async keyword when I refactored to `.then`, will update.

frontend/assets/stylesheets/averta.css
Ian-MacLeod: This seemed to work, but I'll remove it anyways if there's a better way.

markrawls: It will work, but it's not optimal for static rendering which is a big thing we get out of Next

frontend/components/AuthBase.js
```diff
@@ -0,0 +1,144 @@
...
+      email: email.toLowerCase(),
+      password,
+      next,
+    })
+      .then(handleSuccess)
+      .catch(handleError);
+  };
+
+  return (
+    <div>
```
markrawls, line 72: Does this _need_ to be a `div` or will a fragment do? If at all possible we should avoid having `div` wrappers so that the parent component can choose how to flow the content.

frontend/components/AuthBase.js
```diff
@@ -0,0 +1,144 @@
...
+          size="large"
+          disabled={processing}
+        >
+          {submitText}
+        </Button>
+      </Form>
+      <Spacer v={16} />
+      {children}
+      <Spacer v={24} />
+      <Grid
```
markrawls, line 108: PLEASE add vertical whitespace between JSX blocks in here, it's much harder to read and navigate without.

frontend/components/AuthBase.js
Ian-MacLeod: I don't see any reason why it needs to be a div, updated.

frontend/components/AuthBase.js
Ian-MacLeod: Added some newlines, let me know what you think.

PR #35, opened 2020-04-15
Use willing-zg plugin for deployment

markrawls:


frontend/package.json
```diff
@@ -1,45 +1 @@
...
...true, "scripts": {"dev": "node next-server.js", "build": "next build", "start": "npm run build && NODE_ENV=production node next-server.js", "pretty-lint": "pretty-quick --check --branch", "eslint": "node_modules/.bin/eslint --ext .js,.jsx ."}, "dependencies": {"@bequestinc/wui": "^1.0.12", "@material-ui/core": "^4.9.10", "@material-ui/icons": "^4.9.1", "axios": "^0.19.2", "express": "^4.17.1", "express-http-proxy": "^1.6.0", "js-cookie": "^2.2.1", "next": "9.3.2", "next-compose-plugins": "^2.2.0", "next-svgr": "^0.0.2", "prop-types": "^15.7.2", "react": "16.13.0", "react-dom": "16.13.0"}, "devDependencies": {"@testing-library/dom": "^7.2.1", "@testing-library/react": "^10.0.2", "babel-eslint": "^10.1.0", "babel-plugin-module-resolver": "^4.0.0", "be-pretty": "^0.9.4", "eslint": "^6.8.0", "eslint-config-airbnb": "^18.1.0", "eslint-config-prettier": "^6.10.1", "eslint-import-resolver-babel-module": "^5.1.2", "eslint-plugin-import": "^2.20.2", "eslint-plugin-jest": "^23.8.2", "eslint-plugin-jsx-a11y": "^6.2.3", "eslint-plugin-prettier": "^3.1.3", "eslint-plugin-react": "^7.19.0", "eslint-plugin-react-hooks": "^3.0.0", "jest": "^25.3.0", "prettier": "^2.0.4", "pretty-quick": "^2.0.1"}}
```
Ian-MacLeod, line 1: I guess this doesn't get auto-formatted, but can we format this with prettier?

backend/backend/settings/zygoat_settings.py
```diff
@@ -25,7 +25,7 @@ def prod_required_env(key, default, method="str"):
 
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = env.str(
+SECRET_KEY = prod_required_env(
     "DJANGO_SECRET_KEY", default="9a=5%_$0cykzvckso!3wo-1mu#&*t$4ur!xlybxx=l_8#zdex5"
```
ambikads, line 29: Unrelated to the code changes. What is the purpose of this secret key? And I am guessing on production, its different from the default value set here?

Ian-MacLeod: It's used by Django - most significantly for signing cookies and tokens generated by the default Django generator. It definitely should be set to something other than this default in production.

ambikads: How is this set in production?

Ian-MacLeod: It's pulled from the environment variables which are set by AWS configuration.

frontend/package.json
markrawls: Sure. Updated.

PR #37, opened 2020-04-16
Preserve query params

Ian-MacLeod:
To be merged after #34. Only the last two commits are part of this PR.


cviveiro:
LGTM, will get on reviewing that other PR now

frontend/components/Login.js
```diff
@@ -17,7 +17,7 @@ const Login = () => {
       </Typography>
       <Typography>
         Forgot your password?&nbsp;
-        <KeepParamsLink href="/password-reset">
+        <KeepParamsLink href="/reset_password">
```
markrawls, line 20: Page paths should be hyphenated, as is standard convention

PR #40, opened 2020-04-20
Add in change email page, endpoint

### LINKED ISSUE #14: Add in `change-email` backend view
Will be pretty similar to the `change-password` view that's currently under review, and a similar endpoint is also in the willing app.

cviveiro:
Closes #14

This does require some token support that we're currently working on, so I've added in some workarounds to make it testable and working for now. It is obviously not ideal as a long term solution but I've got the plugin PR out and Ian is working on making that more flexible, so hopefully we'll be able to replace soon, maybe even before this gets merged.

I do also think we should send a confirmation email and only change once we get a hit to the link from the confirmation email, like seems standard practice these days. I'll get to work on adding that in later today hopefully but in case other things come up, this will fulfill minimum functionality we'll need.


backend/authentication/views.py
```diff
@@ -68,6 +68,23 @@ def change_password(drf_request):
     return check_and_change_password(drf_request, user, new_password)
 
 
+@api_view(["POST"])
+@permission_classes([IsAuthenticated])
```
cviveiro, line 72: I figured this was enough, but the change password view above also needs the current password and checks to see whether that matches things, adding another layer of security. Do we want that in here? If not, do we want to remove that in the password change?

Ian-MacLeod: I think that if we are going to allow email changes, we should password protect it. We don't want to allow people to change the email and hijack the account if someone leaves their account logged in.

backend/authentication/views.py
```diff
@@ -68,6 +68,23 @@ def change_password(drf_request):
...
+def change_email(drf_request):
+    user = drf_request.user
+    new_email = drf_request.data.get("new_email")
+
+    # TODO get a confirmation email going with a link that will
+    # verify the email address.
+    try:
+        user.email = new_email
+    except ValidationError:
+        return make_response(False)
```
Ian-MacLeod, line 82: This assignment will never trigger a ValidationError. I think the nicest approach here would be to create an EmailChangeSerializer (I think there should probably be a UserSerializer that this, LoginSerializer, and RegisterSerializer inherit from, but maybe that can be handled in a separate PR) and use it here. We could even use a DRF GenericView (we're not using generic views for login/registration because drf views don't have csrf protection on unauthenticated endpoints).

frontend/pages/change_email.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
Ian-MacLeod, line 1: We prefer hyphens in urls: change the file name to `change-email.js`

backend/authentication/views.py
cviveiro: Added and tested (by removing the email restriction on the frontend form), we catch invalid email addresses with the serializer.

frontend/components/ChangeEmail.js
```diff
@@ -0,0 +1,101 @@
...
+          : 'The email you entered was not valid';
+      setInputErrors({ submitError });
+    }
+  };
+
+  const successMessage = () => {
+    return (
+      <div>
+        <Typography variant="h4"> Change Email</Typography>
+        <body>Your email has been changed!</body>
```
Ian-MacLeod, line 62: The body tag has a special semantic meaning (see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body), so we shouldn't use it here. Another typography tag seems appropriate.

frontend/components/ChangeEmail.js
```diff
@@ -0,0 +1,101 @@
...
+      errors.newEmail = 'Please enter your email address.';
+    }
+
+    setInputErrors(errors);
+
+    return Object.keys(errors).length === 0;
+  };
+
+  const handleSubmit = async e => {
+    e.preventDefault();
```
Ian-MacLeod, line 37: Same comment here as in the other PR about `preventDefault` in async functions.

cviveiro:
Changes made and tested!

frontend/pages/change-email.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
```
markrawls, line 2: Not needed in NextJS, it's implicit and handled by the server renderer.

frontend/components/ChangeEmail.js
```diff
@@ -0,0 +1,108 @@
+import React, { useState } from 'react';
```
markrawls, line 1: `import React from 'react';` is never necessary under NextJS, but `import {useState} from 'react';` is.

backend/authentication/views.py
```diff
@@ -68,6 +68,26 @@ def change_password(drf_request):
...
 
+@api_view(["POST"])
+@permission_classes([IsAuthenticated])
+def change_email(drf_request):
+    user = drf_request.user
+    password = drf_request.data.get("password")
+    new_email = drf_request.data.get("new_email")
+
+    if not user.check_password(password):
+        return make_response(False, {"error": AUTH_FAILURE})
```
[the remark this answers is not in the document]
cviveiro: I found myself getting annoyed with having to put in the `200_OK` and `400_BAD_REQUEST` all the time, which started the function, and it has grown a bit since then. I'm definitely fine with taking out the response data manipulation and putting that in the functions, but I do think there's a case to be made for some reasonable status defaults that are taken care of in a lightweight helper function.

frontend/utils/API.js
```diff
@@ -23,10 +26,30 @@ const login = payload => API.post('auth/login/', payload);
 
 const logout = () => API.post('auth/logout/');
 
-const refresh = () => API.post('auth/token/refresh/');
+const refresh = async () => {
+  try {
+    const response = await API.post('auth/token/refresh/');
+    // TODO this is a holdover until we've got plugin support for tokens for portunus,
+    // remove this once that's available.
+    window.accessToken = response.data.access;
```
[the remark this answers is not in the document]
cviveiro: So I can definitely change to not break for this PR, but this is a placeholder until we get the token `willing-zg` plugin going and we get a change in to define which components are "protected" and which aren't. With the plugin, we'll have a singleton object that will fetch and keep track of the access token that we'll use here. Does that sound ok? Ian and I came up with that as an initial solution.

Relevant PR here: https://github.com/MetLifeLegalPlans/willing-zg/pull/1

backend/authentication/views.py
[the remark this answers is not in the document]
markrawls: The `200_OK` is the default if the status is not provided.

I suppose my core complaints are twofold:
1. The usage of a raw HTTPResponse instead of the built-in extremely robust DRF Response
2. The `success` field in the output, which is just duplicating information we already know.

If we want the standard error from this function to `HTTP_400_BAD_REQUEST` then so be it, but IMHO the function should be less involved than it is.

```python
from rest_framework.response import Response
from rest_framework import status

def response(*args, error=False):
    return Response(*args, status=status.HTTP_400_BAD_REQUEST if error else None)
```

Or even just creating our own `Box` around `status`

```python
# status.py
from rest_framework import status

codes = Box({
    400: status.HTTP_400_BAD_REQUEST,
    404: status.HTTP_404_NOT_FOUND,
    # etc
})

###################

# views.py
from status import codes

@api_view()
def my_view(request):
    return Response({'error': 'You messed up somewhere down the line'}, status=codes.400)
```

frontend/utils/API.js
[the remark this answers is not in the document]
markrawls: I merged and released that plugin, so that should be set.

Basically my only complaint with that is that the `isLoggedIn` function does not closely relate to the user's actual state of authentication, so I don't think we should use the one in that module right now.

...also now that I look at it `if (this.fetchToken())` will always return `true` in that context, because that's an async function and promises are truthy but that's not necessarily a problem upfront.

That said, seems more than sufficient for an initial implementation.

backend/authentication/views.py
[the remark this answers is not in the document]
Ian-MacLeod: I agree that DRF API views are nicer to use and that the success key in the response doesn't really make sense. I will say that for the login/registration views specifically, we do want to use vanilla Django views because DRF doesn't enforce CSRF protection on unauthenticated views. I think a separate PR to convert the not login/registration Django views to DRF views would make sense - I'll make an issue.

I'm not necessarily opposed to a thin wrapper around the DRF Response, but I'm not sure that there is much to be gained by it either at this point.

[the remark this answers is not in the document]
cviveiro: I'm arguing for the first example in Mark's comment. Super lightweight, just taking care of the statuses if you want to use the defaults.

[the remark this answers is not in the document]
markrawls: Yep, I'm all on board for that one

cviveiro:
Changes made and tested.

Thanks for the review @markrawls! I'll make the same changes to the other open PR and then I'll make another one where I make these changes to some of the other code I've written here that can use similar changes.

cviveiro:
Rebased

cviveiro:
Looks like linting errors are from social auth stuff:

```
yarn run v1.22.0
$ node_modules/.bin/eslint --ext .js,.jsx .

/code/components/SocialAuth/index.js
  2:25  error  Unable to resolve path to module 'react-google-login'                                     import/no-unresolved
  3:27  error  Unable to resolve path to module 'react-facebook-login/dist/facebook-login-render-props'  import/no-unresolved
```

cviveiro:
Yeah looks like, how would I fix that?

Ian-MacLeod:
Rebuilding the containers should do it - bring them down, delete the existing containers using `docker system prune -a --volumes` (if you have other stopped containers you don't want to delete then you can use `docker rm` instead) and then bring them back up with  `docker-compose up`

PR #42, opened 2020-04-21
Add in change password page

cviveiro:
This is based off of #40 to include the token changes that this also needs to work. Only the last commit is relevant to this PR. Happy to close and wait on the other if desired to keep things cleaner.

This should cover all of the pages we expect to be in the first real iteration of Portunus I think, along with other PRs that are out.

Tested, works as expected.


frontend/utils/constants.js
```diff
@@ -0,0 +1,4 @@
+const invalidPasswordError = 'invalid_password';
+const authFailure = 'auth_failure';
```
Ian-MacLeod, line 2: I think I'd slightly prefer SCREAMING_SNAKE_CASE for these constants.

frontend/components/ChangePassword.js
```diff
@@ -0,0 +1,111 @@
...
+      errors.newPassword2 = 'The two passwords you entered do not match.';
+    }
+
+    setInputErrors(errors);
+
+    return Object.keys(errors).length === 0;
+  };
+
+  const handleSubmit = async e => {
+    e.preventDefault();
```
Ian-MacLeod, line 41: @markrawls raises a good point [here](https://github.com/MetLifeLegalPlans/portunus/pull/34#discussion_r412428178) about preventDefault in async functions. Possible solutions: use Promise.then instead of async/await, or refactor the async part of the callback to a helper function.

frontend/components/ChangePassword.js
```diff
@@ -0,0 +1,111 @@
...
+          : 'Your current password did not match the one we have on file. Try again.';
+      setInputErrors({ submitError });
+    }
+  };
+
+  const successMessage = () => {
+    return (
+      <div>
+        <Typography variant="h4">Change Password</Typography>
+        <body>Your password has been changed!</body>
```
Ian-MacLeod, line 65: Use something other than `body` here

cviveiro:
Changes made and tested, thanks for the review!

cviveiro:
This is all rebased and ready.

PR #43, opened 2020-04-21
Add testing

Ian-MacLeod:
Little bit of a PR pileup, going to request a review from @markrawls on #34. In the meantime, only the last commit here is part of this PR.

### DIFF
--- a/backend/authentication/factories.py
+++ b/backend/authentication/factories.py
@@ -1,0 +1,13 @@
+import factory
+
+from authentication.models import User
+
+
+class UserFactory(factory.DjangoModelFactory):
+    class Meta:
+        model = User
+
+    email = factory.Faker("email")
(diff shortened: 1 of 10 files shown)


Ian-MacLeod:
Some more organization sounds good -- We can organize tests into classes with pytest, too.

Ian-MacLeod:
There was an issue with codebuild after the repo org was switched, but tests should be running now. Thoughts on the new structure? I reorganized the methods into classes - might be possible to move them to different files or simplify them further using a base class, but I think this will do for now.

cviveiro:
SOO much better, thanks!! I definitely don't mind them in the same file as long as they're split out some other way, and I like the way you split them into view tests and then also took the utility functions out.

LGTM

PR #47, opened 2020-04-27
Keep track of user authentication status and protect private pages

Ian-MacLeod:
The ProtectedPage component will redirect to the login page if the user is not authenticated.

There aren't currently any pages to protect, but I set the homepage to be a protected page and added a logout button for testing.


frontend/components/AuthBase.js
```diff
@@ -11,16 +12,22 @@ import Typography from '@wui/basics/typography';
 import TabDivider from '@wui/layout/tabDivider';
 
 import SocialAuth from '@@/components/SocialAuth';
-import useInputFieldState from '../utils/hooks';
+import { useGlobalContext, useInputFieldState } from '../utils/hooks';
```
cviveiro, line 15: While you're in here, should change to `@@/utils/hooks`? has seemed like we're trying to move to all using that rather than the `..`, though that might not be the case and i might have misunderstood

Ian-MacLeod:
Hmm, I agree the protected page seems like a good default. I'll see if there is a nice way to achieve that. I also agree that the protected page pattern is one that we will probably want in other apps, but I think Portunus is a bit of a special case. Adding something to willing-zg makes sense to me.

Ian-MacLeod:
This is the best that I could think of for having components be protected by default. To mark a page as public we set `public` to `true` on the component.

cviveiro:
I like that interface. 

Looks reasonable to me but I do think someone else should take a look at this if it's not holding anything up. If it is we should get it in.

PR #51, opened 2020-05-08
Create user flow

Ian-MacLeod:
This allows administrators to create new accounts. After the account is created, the user gets an email with a link that lets them set a password for the account.


backend/authentication/serializers.py
```diff
@@ -77,3 +78,16 @@ class ChangeEmailSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
         fields = ("email",)
+
+
+class AdminCreateUserSerializer(serializers.ModelSerializer):
```
cviveiro, line 83: would prefer `CreateAdminUserSerializer` but don't feel strongly about it

Ian-MacLeod: Hmm, the serializer is for admins to create normal users, not to create users with admin privileges. Maybe just `CreateUserSerializer` is best?

cviveiro: Ahh I'd misunderstood, yeah I think `CreateUserSerializer` is good and more clear!

Ian-MacLeod:
rebased and updated to use the new PortunusMailer class

PR #56, opened 2020-05-16
EMAIL: add email component

kborer:
Ran the `willing-zg` plugin, made some slight modifications, and migrated the current email logic.


kborer:
I tested it locally and saw the email printed to the console.

PR #57, opened 2020-05-18
Remove data manipulation from response utility

### LINKED ISSUE #45: Convert vanilla Django views to use DRF's api_view
We can also get rid of `authentication.utils.make_response` while we're at it.

See discussion here: https://github.com/MetLifeLegalPlans/portunus/pull/40#discussion_r413407202

We will want to keep login/register/social_auth views as Django views to maintain CSRF protection.

laks12metlife:
Closes #45 

Removed the response data manipulation.  Tested the case where user enters an invalid password during account creation and worked fine.

Attempted to use DRF Response like so:

```
from rest_framework.response import Response
...
response = Response(json.dumps(data), content_type="application/json", status=status)
```
but didn't work.  When testing with invalid password, returned 500 error.


backend/authentication/utils.py
```diff
@@ -29,13 +29,7 @@ def make_response(success, data=None, status=None):
...
         status = status_codes.HTTP_200_OK if success else status_codes.HTTP_400_BAD_REQUEST
 
-    response_data = {"success": success}
-    if data:
-        response_data.update(data)
-
-    response = HttpResponse(
-        json.dumps(response_data), content_type="application/json", status=status
-    )
+    response = HttpResponse(json.dumps(data), content_type="application/json", status=status)
```
Ian-MacLeod, line 32: I think it would make sense to default success to True so that we don't have to pass a redundant value for success when we are supplying a specific status.

laks12metlife: ok, that makes sense

Ian-MacLeod:
As far as switching to DRF responses goes - We would need to avoid using the make_response util in vanilla django views (which is just the login/register views).

I don't see a problem with merging this for now, though.

laks12metlife:
rebased with master

backend/authentication/utils.py
```diff
@@ -25,17 +25,11 @@
 FROM_EMAIL = "Willing <hello@willing.com>"
 
 
-def make_response(success, data=None, status=None):
+def make_response(success=True, data=None, status=None):
```
cviveiro, line 28: nit: could we change the functions in `views.py` that are just calling `make_response(True)` to use the new default? i.e. they won't need to pass in anything and just do `make_response()`

cviveiro:
Needs to be updated with master.
```
git checkout master && git pull [willing remote] master
git checkout [your branch] && git rebase master
```

laks12metlife:
Ok, this branch is updated with master.  Ian, once the approval is done, do you want me to merge into master or will somebody else take care of that?

PR #60, opened 2020-05-20
Change the route for reset password to use dash instead of underscore

laks12metlife:
Changed the underscore to dash in the password reset URL and the corresponding JS folder. Tested reset password locally and worked fine.  Changing for consistency and bc dashed are preferred in urls.


laks12metlife:
sorry, a rebase was needed

PR #61, opened 2020-05-22
Allow redirects to willing subdomains

Ian-MacLeod:
Noticed that this wasn't in yet when checking out dev.members.willing.com.

We probably want this to be configurable - We could use environment variables, but maybe the Django admin is a better place for that.

### DIFF
--- a/backend/backend/settings/__init__.py
+++ b/backend/backend/settings/__init__.py
@@ -68,7 +68,10 @@ AUTH_PASSWORD_VALIDATORS = [
 DEFAULT_REDIRECT_URL = prod_required_env(
     "DJANGO_DEFAULT_REDIRECT_URL", "http://localhost:3000"
 )
-VALID_REDIRECT_HOSTNAMES = ["localhost"]
+VALID_REDIRECT_HOSTNAMES = ["*.willing.com"]
+
+if not PRODUCTION:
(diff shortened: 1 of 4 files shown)


backend/backend/settings/__init__.py
```diff
@@ -68,7 +68,10 @@
 DEFAULT_REDIRECT_URL = prod_required_env(
     "DJANGO_DEFAULT_REDIRECT_URL", "http://localhost:3000"
 )
-VALID_REDIRECT_HOSTNAMES = ["localhost"]
+VALID_REDIRECT_HOSTNAMES = [".willing.com"]
```
cviveiro, line 71: Hmm. I think I'd prefer another way of declaring that all subdomains are valid rather than a `.` at the beginning. Maybe make it more like a wildcard? With something like `*.willing.com`, it's much clearer what's going on. Also, bonus, you could use something like [fnmatch](https://docs.python.org/3.7/library/fnmatch.html) that would do the matching and also allow for some more advanced domain defining out of the box.

Ian-MacLeod: There is some precedent in the Django settings for using a string starting with a period to allow subdomains. I think I'm OK with using fnmatch instead as long as we are careful about the patterns that we use.

Ian-MacLeod: Updated

PR #66, opened 2020-06-04
Run zg update, add new token logic

cviveiro:
The only manual changes are in `frontend/API.js`, to work with the promise that's now returned by the token fetcher. Portunus needed an unauthenticated API call to make the actual token request (can't make a token request while waiting for the token fetcher to get one...) and decided to use that for the other unauthenticated endpoints.

Tested, works.


frontend/utils/API.js
```diff
@@ -3,6 +3,15 @@ import Cookies from 'js-cookie';
...
 
+const unauthenticatedAPI = axios.create({
+  baseURL: '/api/',
+  responseType: 'json',
+  headers: {
+    'X-REQUESTED-WITH': 'XMLHttpRequest',
+    'Content-Type': 'application/json',
+  },
+});
+
```
Ian-MacLeod, line 14: Since it's the same, we should probably store the configuration object in a variable and use it for both unauthenticatedAPI and API.

cviveiro:
Ok, broke up into two commits, one purely for the `zg update` (though i'll want them squashed on merge because things don't work when they're not together), made the requested change, and retested. Works as expected.

Sorry I forgot about this one for a bit >.<

PR #72, opened 2020-06-20
DOCKER: Add configuration to build production image

ambikads:
The dev Dockerfile is renamed to `Dockerfile.dev`. 
Non dev builds will be built from `Dockerfile`

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -2,7 +2,12 @@ FROM node:13
 
 RUN mkdir /code
 WORKDIR /code
-ADD package.json /code/
+
+COPY package.json yarn.lock ./
 RUN yarn install
 
-CMD yarn dev -p 3000
+COPY . .
+RUN npm run build
+
+EXPOSE 3000
+CMD npm start -p 3000
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -7,7 +7,6 @@ COPY package.json yarn.lock ./
 RUN yarn install
 
 COPY . .
-RUN npm run build
 
 EXPOSE 3000
-CMD npm start -p 3000
+CMD yarn start -p 3000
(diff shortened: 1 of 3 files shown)


frontend/Dockerfile
```diff
@@ -2,7 +2,12 @@ FROM node:13
...
+
+COPY package.json yarn.lock ./
 RUN yarn install
 
-CMD yarn dev -p 3000
+COPY . .
+RUN npm run build
+
+EXPOSE 3000
+CMD npm start -p 3000
```
markrawls, line 13: Switching between `yarn` and `npm` in the middle of the Dockerfile is a bit suboptimal

frontend/Dockerfile
```diff
@@ -2,7 +2,12 @@ FROM node:13
...
 RUN mkdir /code
 WORKDIR /code
-ADD package.json /code/
+
+COPY package.json yarn.lock ./
 RUN yarn install
 
-CMD yarn dev -p 3000
+COPY . .
+RUN npm run build
```
markrawls, line 10: This belongs in `CMD` - `CMD yarn build && yarn start -p 3000`

frontend/Dockerfile
markrawls, line 13: If I remember correctly, EB will choose a port dynamically at runtime and put it in `$PORT`?

markrawls:
This probably belongs in a `willing-zg` plugin

frontend/Dockerfile
ambikads: `yarn start` includes `build`

frontend/Dockerfile
ambikads: We need to expose a port for docker platform

ambikads: done

ambikads:
@markrawls addressed comments

PR #75, opened 2020-06-23
Add in login settings page

### LINKED ISSUE #50: login settings page
To minimize the need for integrations, the current members design will send users to portunus to update account email and password. Just like sign-in, we'll need a way for Portunus to bounce the user back to the referring application after they are done and click "back". Here is the design:

login_details

cviveiro:
Closes #50 

First go at this. Right now, for the sake of quickness, I made it so that this page just redirects to the change-email and change-password pages. This way was just easier. I am starting to think that this is the right way to move forward anyways. We'll have a lot of settings for mfa and we'll need to add some for social auth, and i think so as to keep this page less cluttered, it might be best to redirect to individual change pages for the different settings. I could be persuaded otherwise though. The settings just aren't as simple to change as they are for the members settings stuff.

Discussion question: this borrows a lot of code from members because it's made to look like the members code, and also the members code has some nice components built out. Is there a good way to share this code going forward? I do think that we'll want to think about it. I'll bring it up in the members standup today.

Two things I think will need changing but I'm not sure how:
1) to get the back button on every page, I added it to `_app.js` and it's ok. Is there a better place to put it? I figure on the change components, since we're in a small form, we can't put it at the top of the screen and it won't be as continuous, but i can move it to individual components if necessary.
2) the `Page` component I'm sure doesn't have the correct css style for it's height. Without that, it was stopping right below the textbox and it was jarringly in the center of the screen, now it reaches my full screen but it's obviously not really scale-able.

Tested locally.


Ian-MacLeod:
I think the back button would be better off inside the Page component (not sure, but that makes more sense to me). We probably don't want it on the login and register pages, right?

For the height issue, we could do something similar to what members does: _app imports a global.css file that sets the html/body/next root elements to fill the page even when the content is smaller. As an alternative, we could use min-height: 100vw or something like that, but I think that's less nice.

backend/authentication/urls.py
```diff
@@ -10,6 +10,7 @@
     path("login/", views.login, name="login"),
     path("social-auth/", views.social_auth, name="social_auth"),
     path("logout/", views.logout, name="logout"),
+    path("users/current/", views.get_current_user_settings, name="get_current_user"),
```
Ian-MacLeod, line 13: I think I like "user/settings/" better as the path for this.

backend/authentication/views.py
```diff
@@ -155,6 +155,13 @@ class RetrieveUserView(RetrieveAPIView):
     queryset = User.objects.all()
 
 
+@api_view(["GET"])
+@permission_classes([IsAuthenticated])
+def get_current_user_settings(request):
+    # TODO add social auth settings and link in the frontend.
+    return make_response(data={"email": request.user.email})
```
Ian-MacLeod, line 162: Fine for now, I suppose, but should probably be handled by a serializer when we add more fields.

frontend/components/BackButton.jsx
```diff
@@ -0,0 +1,40 @@
...
+import Button from '@wui/input/button';
+import Typography from '@wui/basics/typography';
+
+const useStyles = makeStyles(theme => ({
+  withPadding: {
+    padding: theme.spacing(2),
+  },
+}));
+
+const BackButton = ({ noPadding }) => {
```
Ian-MacLeod, line 16: Do we need the noPadding prop here? If we don't intend to use it I'd favor getting rid of it.

frontend/components/SettingRow.jsx
```diff
@@ -0,0 +1,79 @@
...
+    width: 100,
+  },
+}));
+
+const AccountSettingRow = ({ value, label, routeTo, hidden }) => {
+  const router = useRouter();
+  const classes = useStyles();
+
+  // TODO maybe get the length of the password from the backend (we'd have to store it)
+  // so that we can properly fill in the password value's length?
```
Ian-MacLeod, line 35: I think this is not necessary (and not desirable IMO)

frontend/components/SettingRow.jsx
```diff
@@ -0,0 +1,79 @@
...
+      [0, 1, 0, 0, theme.palette.custom.border],
+      [0, 1, 0, 0, theme.palette.custom.border, 'inset'],
+    ],
+  },
+  edit: {
+    color: theme.palette.blue.hover,
+  },
+  update: {
+    width: 100,
+  },
```
Ian-MacLeod, line 27: Is this used?

frontend/components/LoginSettings.jsx
```diff
@@ -0,0 +1,77 @@
...
+const useStyles = makeStyles(theme => ({
+  root: {
+    padding: [[32, 20]],
+    [theme.breakpoints.notPhone]: {
+      padding: [[32, 60]],
+    },
+  },
+}));
+
+const LoginSettingsForm = () => {
```
Ian-MacLeod, line 26: This is a nit, but I think I like matching the name of the file with the name of the default exported component.

Ian-MacLeod:
> Discussion question: this borrows a lot of code from members because it's made to look like the members code, and also the members code has some nice components built out. Is there a good way to share this code going forward? I do think that we'll want to think about it. I'll bring it up in the members standup today.

One way is to create/update a wui component if something is general enough. Might want to do that for the `TableHeader` component, for example.

cviveiro:
That global css stuff is magic.

Thanks for taking a look so quickly!!! Changes made and tested. I'll look into making a wui PR with the `TableHeader` component.

frontend/components/Page.jsx
```diff
@@ -0,0 +1,36 @@
+import PropTypes from 'prop-types';
+import { observer } from 'mobx-react';
+import Paper from '@material-ui/core/Paper';
+import { makeStyles } from '@material-ui/core/styles';
+import classNames from 'classnames';
+
+import BackButton from '@@/components/BackButton';
+
+// TODO how to get this to look good in relative terms (%'s)
```
Ian-MacLeod, line 9: This comment seems unnecessary now

cviveiro: Thanks for catching, removed

PR #76, opened 2020-06-25
Add logout path

Ian-MacLeod:
It will be useful for other apps to have a path to redirect to rather than an API endpoint to log users out.

### DIFF
--- a/backend/backend/views.py
+++ b/backend/backend/views.py
@@ -1,10 +1,14 @@
 from django.http import HttpResponse
 from django.views.decorators.http import require_GET
 from django.views.decorators.csrf import ensure_csrf_cookie
+from django.contrib.auth import logout as logout_user
+from django.shortcuts import redirect
 from rest_framework.decorators import api_view, permission_classes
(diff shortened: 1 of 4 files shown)


backend/backend/views.py
```diff
@@ -18,3 +22,10 @@ def is_authenticated(request):
 @require_GET
 def set_csrf(request):
     return HttpResponse("success", status=HTTP_200_OK)
+
+
+def logout(request):
```
cviveiro, line 27: We have this already and it's an endpoint that's exposed (https://github.com/MetLifeLegalPlans/portunus/blob/a1210e99c66ed1921c389ccfe0e79cacf89d970a/backend/authentication/views.py#L57) except that it doesn't redirect. Do we still need the other one? If we do, any way we can not have the two endpoints and combine?

Ian-MacLeod: I was thinking that this one would supersede the existing one. I didn't delete the old one here to avoid breaking log out for members. Was planning to 1. release this, 2. update members to use this, 3. delete the old view.

PR #80, opened 2020-07-01
AUTH: return uuid in new user view error response

### LINKED ISSUE #71: AUTH: return uuid in new user view error response
When the new user endpoint `/api/auth/users/`, `CreateUserView`,  responds with an error because the email already exists, we want to include the existing user's uuid in the response so the member app can look up the member with the existing email for the admin panel.

swang192:
Close #71


swang192:
Don't review, will be changing the logic soon to better fit frontend needs

swang192:
Ready for review, not sure why but the response isn't camel-casing the object keys

backend/authentication/views.py
```diff
@@ -146,6 +146,21 @@ class CreateUserView(CreateAPIView):
...
     serializer_class = CreateUserSerializer
 
+    def create(self, request, *args, **kwargs):
+        serializer = self.serializer_class(data=request.data)
+
+        if serializer.is_valid():
+            user = serializer.create(request.data)
+            return make_response(True, {"portunus_uuid": str(user.portunus_uuid)})
+        else:
+            existing_user = User.objects.get(email=request.data["email"])
```
Ian-MacLeod, line 156: We shouldn't assume that this user necessarily exists here, so we want to use `User.objects.first` here.

backend/authentication/views.py
```diff
@@ -146,6 +146,21 @@ class CreateUserView(CreateAPIView):
...
+            data = serializer.errors
+
+            if serializer.errors["email"] and existing_user:
```
Ian-MacLeod, line 159: The serializer email error check doesn't hurt, but feels redundant here because an existing user is a problem no matter what.

backend/authentication/tests/views/test_views.py
```diff
@@ -111,3 +114,44 @@ def test_logout(self, post, client, authenticate_and_test):
...
+        request = factory.post(
+            "/user/", {"email": new_email, "password": VALID_PASSWORD,}, format="json"
+        )
+        force_authenticate(request, user=admin_user)
+        response = view(request)
+        new_user = User.objects.get(email=new_email)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            json.loads(response._container[0])["portunus_uuid"], str(new_user.portunus_uuid)
```
Ian-MacLeod, line 138: I think `json.loads(response._container[0])` can be replaced by `response.json()`. In general it's best to avoid properties marked with the leading underscore when possible (they are intended to be private).

backend/authentication/tests/views/test_views.py
```diff
@@ -111,3 +114,44 @@ def test_logout(self, post, client, authenticate_and_test):
...
+        view = CreateUserView.as_view()
+        request = factory.post("/user/", USER_DATA, format="json")
+        force_authenticate(request, user=admin_user)
+        response = view(request)
+
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(
+            json.loads(response._container[0])["portunus_uuid"], admin_user.portunus_uuid
+        )
+        self.assertEqual(json.loads(response._container[0])["user_exists"], True)
```
Ian-MacLeod, line 157: Good time to use `assertTrue`

backend/authentication/tests/views/test_views.py
swang192: ahh okay, I'll try it. Wasn't sure how to access it otherwise

swang192: `response.json()` didn't work, but I did some digging and found this:

Screenshot from 2020-07-09 20-01-32

Tried this, and not sure why, but render couldn't be called on response. But then I tried doing `response.content` directly, and that ended up working, so changed it to that instead.

swang192:
comments addressed

backend/authentication/tests/views/test_views.py
Ian-MacLeod: Ah, perhaps the `json` method is only available when using the test client. `.content` seems like a fine solution anyways. I think the render method that those docs mention might be specifically required if the response is rendered using templates.

backend/authentication/tests/views/test_views.py
```diff
@@ -111,3 +114,44 @@ def test_logout(self, post, client, authenticate_and_test):
...
+        admin_user = UserFactory(**USER_DATA)
+        admin_user.is_staff = True
+        admin_user.is_superuser = True
+        admin_user.save()
+
+        factory = APIRequestFactory()
+        view = CreateUserView.as_view()
+        new_email = "newuser@test.com"
+        request = factory.post(
+            "/user/", {"email": new_email, "password": VALID_PASSWORD,}, format="json"
```
Ian-MacLeod, line 130: I missed this on the first look, but it would be best to get the url using `django.urls.reverse` rather than passing a hard-coded string path.

PR #88, opened 2020-07-08
zg update

kborer:
related to https://github.com/MetLifeLegalPlans/members/pull/241


backend/shared/mailer.py
```diff
@@ -1,17 +1,28 @@
 from django.core.mail import EmailMultiAlternatives
 from django.template.loader import render_to_string
 
-from backend.settings import SUPPORT_PHONE_NUMBER, SUPPORT_EMAIL_ADDRESS
+from backend.settings import SUPPORT_PHONE_NUMBER, SUPPORT_EMAIL_ADDRESS, PANEL_EMAIL_ADDRESS
```
Ian-MacLeod, line 4: This is a change that should be made in willing-zg, but we should prefer importing from django.conf.settings.

ambikads:
I think this also didn't include update from https://github.com/MetLifeLegalPlans/willing-zg/pull/20

PR #93, opened 2020-07-13
FRONTEND: add nav bar

### LINKED ISSUE #91: add nav bar to login-settings
make it look like the members app, so when members are redirected to update their account settings it is not visually jarring.

Screenshot from 2020-07-12 23-56-05
Screenshot from 2020-07-12 23-55-58

swang192:
Close #91. Also changes the logo to the following: 
image

Screenshot from 2020-07-22 13-02-56
image


frontend/components/Nav.jsx
```diff
@@ -0,0 +1,103 @@
...
+  const handleClose = () => {
+    toggleMenu();
+    setMenuAnchor(null);
+  };
+
+  return (
+    <AppBar position="sticky" className={classes.appBar} elevation={0}>
+      <Toolbar className={classes.toolBar}>
+        <Grid container direction="row" justify="space-between" alignItems="center">
+          <Link href="/">
```
cviveiro, line 68: Do we want this to be different for Portunus? This page isn't great... what about account settings?

frontend/utils/hooks.js
```diff
@@ -1,4 +1,4 @@
-import { useState, useContext } from 'react';
+import { useState, useContext, useEffect } from 'react';
```
cviveiro, line 1: unused import?

swang192:
This was supposed to be a quick thing for the demo, but I guess since it's past demo time, I'll go polish this off some more

Ian-MacLeod:
> Do we want this to be different for Portunus? This page isn't great... what about account settings?

Seems good to me. Also, the current `/` page was originally added as a placeholder - I think that it would make sense if account settings were at `/`  instead (probably not something to do in this pr).

cviveiro:
> the current `/` page was originally added as a placeholder - I think that it would make sense if account settings were at `/` instead (probably not something to do in this pr).

Ahhh yeah, that sgtm in another PR

cviveiro:
@swang192 if you have too much on your plate, lmk and I can finish this up.

swang192:
Ready for review again, added a bunch of hooks and stuff from members app. Menu just has log out in it for now.

PR #94, opened 2020-07-14
REGISTRATION: add password confirmation field

kborer:
Screenshot from 2020-07-14 19-01-12


frontend/components/AuthBase.js
```diff
@@ -11,9 +11,16 @@ import Typography from '@wui/basics/typography';
 
 import { useGlobalContext, useInputFieldState } from '@@/utils/hooks';
 
-const AuthBase = ({ submitCredentials, submitText, headerText, children }) => {
+const AuthBase = ({
+  submitCredentials,
+  submitText,
+  headerText,
+  confirmPassword: showConfirmPassword,
```
cviveiro, line 18: Nit: feels like this is misnamed because it also deals with showing the password help. So it's more generally related to registration. Perhaps `isForRegistration` or `showRegistrationDetails` or something similar would be more illustrative.

kborer: I thought the same thing at first, but I think it is okay because a user will only need the help text if they are entering a new password, and they will always get the confirmPassword field when they are entering a new password.

PR #98, opened 2020-07-16
CHANGE EMAIL: send confirmation email before actually changing user's…

### LINKED ISSUE #41: Send confirmation email before actually changing a user's email
Working off of #14 to make more robust.

laks12metlife:
… email

Closes #41 

Couple notes:
- Occasionally I will get a 401 response from the call to /auth/change-email/complete/.  I will consistently get a 401 error and if I bring down the containers and try again later, I will consistently get 200 OK.  I can't figure out the reason why but I feel like it might be something external to these changes.
- In the `CompleteChangeEmail` component I'm using the `Page` component, so this makes the look of everything different from what the reset password flow looks like.  Also, the page that displays whether the email update was successful or not, opens in a new browser window.  So the Back button is useless.  

Tested locally.  The verification email with the link shows up in the terminal window where the docker containers are running, and clicking the link works as expected.


backend/authentication/change_email_token.py
```diff
@@ -0,0 +1,14 @@
+from datetime import timedelta
+
+# from rest_framework_simplejwt.tokens import Token, BlacklistMixin
```
Ian-MacLeod, line 3: rm

backend/authentication/views.py
```diff
@@ -63,16 +64,36 @@ def change_password(request):
...
     new_email = request.data.get("new_email")
 
     if not user.check_password(password):
         return make_response(False, {"error": AUTH_FAILURE})
 
+    if not new_email:
+        return make_response(False)
+
     # TODO get a confirmation email going with a link that will
     # verify the email address.
```
Ian-MacLeod, line 79: rm

backend/authentication/views.py
```diff
@@ -63,16 +64,36 @@ def change_password(request):
...
+def request_email_change(request):
     user = request.user
     password = request.data.get("password")
     new_email = request.data.get("new_email")
 
     if not user.check_password(password):
         return make_response(False, {"error": AUTH_FAILURE})
 
+    if not new_email:
+        return make_response(False)
```
Ian-MacLeod, line 76: I think it would be best to check the new email using Django's `EmailValidator` here, rather than just making sure it exists.

backend/authentication/views.py
```diff
@@ -63,16 +64,36 @@ def change_password(request):
...
+
+@api_view(["POST"])
+@permission_classes([IsAuthenticated])
+def update_email(request):
+    user = request.user
+    token = request.data.get("token")
+    new_email = request.data.get("newEmail")
+
+    if not user:
+        return make_response(False)
```
Ian-MacLeod, line 92: Is this necessary? I would expect that the IsAuthenticated permission class would take care of this.

frontend/components/CompleteChangeEmail.js
```diff
@@ -0,0 +1,75 @@
...
+      <Page>
+        <Container maxWidth="md" className={classes.root}>
+          <Typography variant="h4">Change Email Error</Typography>
+          <Typography variant="body1">{validationError}</Typography>
+        </Container>
+      </Page>
+    );
+  };
+
+  handleNewEmail();
```
Ian-MacLeod, line 66: handleNewEmail should be performed in a `useEffect` hook with empty dependencies to make sure it runs only once. Also, we should probably include some kind of loading state to show while waiting for the API response.

Ian-MacLeod:
I'll try to reproduce the 401 problem - is there anything in particular you did to produce the problem?

> In the CompleteChangeEmail component I'm using the Page component, so this makes the look of everything different from what the reset password flow looks like. Also, the page that displays whether the email update was successful or not, opens in a new browser window. So the Back button is useless.

I think we likely should update the password reset pages to use `Page` also

laks12metlife:
> I'll try to reproduce the 401 problem - is there anything in particular you did to produce the problem?

Ugh, no sorry Ian I really don’t understand why I’m getting the 401 errors.  But, I am pretty certain that in the API call to the backend, the access token is getting set in the Authorization header, so I don’t think it’s that.  Here’s an example of what I see in the log when I get the 401.  You can see that the request to change email goes through successfully.  Then the csrf and access token get set.  Then an unexpected 401 on the attempt to actually update the email. 

```
backend_1   | [19/Jul/2020 15:14:37] "POST /api/auth/change-email/ HTTP/1.1" 200 4
backend_1   | [19/Jul/2020 15:14:42] "POST /api/auth/token/refresh/ HTTP/1.1" 200 396
backend_1   | [19/Jul/2020 15:14:42] "GET /api/set_csrf/ HTTP/1.1" 200 7
backend_1   | Unauthorized: /api/auth/change-email/complete/
backend_1   | [19/Jul/2020 15:14:42] "POST /api/auth/change-email/complete/ HTTP/1.1" 401 51
```

frontend/components/CompleteChangeEmail.js
laks12metlife: Changed to use useEffect hook.

For the loading, I think the ProjectedPage.js already handles the loading state for all the pages.

```
  if (store.loading) {
    return <div>Loading...</div>;
  }
```

Do you want something more here?

backend/authentication/views.py
laks12metlife: Good idea, updated.
The email validation was being handled on the frontend.  Since we are setting type=email in the Textbox component, it does the email validation for us.  I changed the type to "text", switched the email validation to the backend point, and now the error message saying “The email you entered was not valid” appears if the email isn’t valid.

laks12metlife:
One more thing that is stumping me.  In the case where a user logs out before clicking their email confirmation link, I expect to be routed to the login page (with localNext set) before I can see that my email has been successfully changed.  This is not the case.  When I request an email change, log out, then click my email confirmation link, I am taken directly to the "Email Successfully Changed" page without having to log back in.  I assumed that after logout, I would not longer be authenticated.  But I'm seeing that /api/is_authenticated/ returns 200 OK after I log out:

```
backend_1   | [19/Jul/2020 15:05:56] "GET /api/logout/ HTTP/1.1" 302 0
frontend_1  | [ event ] build page: /login
frontend_1  | [ wait ]  compiling ...
frontend_1  | [ ready ] compiled successfully
backend_1   | Bad Request: /api/auth/token/refresh/
backend_1   | [19/Jul/2020 15:05:56] "POST /api/auth/token/refresh/ HTTP/1.1" 400 43
backend_1   | [19/Jul/2020 15:05:56] "GET /api/set_csrf/ HTTP/1.1" 200 7
backend_1   | [19/Jul/2020 15:06:13] "GET /api/is_authenticated/ HTTP/1.1" 200 31
```
Does this have to do with logout returning 302?

frontend/components/CompleteChangeEmail.js
Ian-MacLeod: The ProtectedPage component handles the loading state while the frontend waits for initial user data to load. In this case, we are waiting for an additional API response to completeUserChangeEmail. As things are right now, it looks like it will show the error state while it waits, which isn't ideal - just some text saying Loading... would be preferable.

frontend/components/ChangeEmail.js
```diff
@@ -98,7 +101,7 @@ const ChangeEmail = () => {
         <Form error={inputErrors.submitError} onSubmit={handleSubmit} noMargin>
           <Textbox
             name="new_email"
-            type="email"
+            type="text"
```
cviveiro, line 104: Maybe I missed this from something else, but why this change?

frontend/components/CompleteChangeEmail.js
```diff
@@ -0,0 +1,84 @@
...
+    try {
+      await completeChangeUserEmail({
+        token,
+        newEmail,
+      });
+      setValidationError(null);
+      setSuccess(true);
+    } catch (error) {
+      const errorMsg =
+        'Your email verification link is invalid or expired. You can try requesting a new one.';
```
cviveiro, line 40: Let's make it clear it's an email change link, or like, that your email has not been changed. Maybe just add "Your email has not been changed."?

frontend/components/ChangeEmail.js
cviveiro: Ah I see, this was for testing. Great! Make sure it gets removed in the final version.

laks12metlife: So when I keep text=email on the frontend, it doesn't display the "The email you entered was not valid” message.  It just shows an input error like so: 
Screenshot from 2020-07-20 13-07-01

But if I change the Textbox type=text, and just let the backend do the validation, I see the error messages that we have specified, not just the default that the Textbox component is returning.  
no_email
invalid_email

Both ways seem fine to me.  If we want both the frontend and backend checks for email validation I can change this back to type=email but just wanted to explain why I got rid of it in the first place :)

cviveiro: I think validation on both ends makes sense! The frontend validation is made to be nice for UI :)

frontend/components/CompleteChangeEmail.js
laks12metlife: ohh okay gotcha. Updated to included a loading state while waiting on api response.

backend/authentication/views.py
```diff
@@ -63,16 +65,33 @@ def change_password(request):
...
 
     if not user.check_password(password):
         return make_response(False, {"error": AUTH_FAILURE})
 
-    # TODO get a confirmation email going with a link that will
-    # verify the email address.
+    try:
+        validate_email(new_email)
+    except ValidationError:
+        return make_response(False, {"error": INVALID_EMAIL})
```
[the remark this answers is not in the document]
laks12metlife: Good catch, updated to make sure email isn't already associated with another user.  I also added a new error type to display an error on the front end for this situation, but I get rid of it if you don't feel it's needed.

backend/authentication/utils.py
```diff
@@ -111,3 +111,16 @@ def check_onetime_token(token_str, user):
...
+    if not token_str:
+        return False
+
+    try:
+        # This will verify that the token is valid and hasn't expired.
+        ChangeEmailToken(token_str)
+        return True
+
+    except TokenError:
+        return False
```
Ian-MacLeod, line 126: I somehow missed this earlier, but we need to check that the email in the token matches the user's email here.

laks12metlife: I might be misunderstanding something, but I'm not sure what you mean by "email in the token".  The way I did it is that the new email is in the url, not the token (I could look into changing this if it's not secure).  But regardless, we are updating the user's existing email with the new email so I'm not clear on needs to be checked here.

Ian-MacLeod: Sorry if I wasn't clear here. The user's current email is included in the token and we should check that it matches the email of the account before we update to the new one. Something like this should work:
```
token = ChangeEmailToken(token_str)
if token['email'] != user.email:
    return False
```

laks12metlife:
Hi @Ian-MacLeod 
So I got back to looking into this and nearly everything works if I Copy Paste the URL instead of opening in the PyCharm window.  The one outstanding issue I'm noticing is that in the case where the user logs out before clicking the link, upon login, they are sent to the settings page rather than the change email complete page.  Some debug statements later, I think https://github.com/MetLifeLegalPlans/portunus/blob/63f582f8edd3269ce52e5fc57dbf09b0cd9309f4/frontend/components/AuthBase.js#L35 is overwriting localNext with "/".  Weirdly enough, I'm only experiencing this in Firefox, and not Chromium.

cviveiro:
needs a rebase again, also @Ian-MacLeod did you have a final request that didn't get submitted?

backend/authentication/utils.py
```diff
@@ -154,3 +154,18 @@ def generate_axes_lockout_response(request, credentials):
...
+def check_change_email_token(token_str, user):
+    if not token_str:
+        return False
+
+    try:
+        # This will verify that the token is valid and hasn't expired.
+        token = ChangeEmailToken(token_str)
+        if token["user_id"] != str(user.portunus_uuid):
+            return False
+        return True
```
Ian-MacLeod, line 168:
```
return token["user_id"] == str(user.portunus_uuid)
```

Ian-MacLeod:
> needs a rebase again, also @Ian-MacLeod did you have a final request that didn't get submitted?

Whoops, not sure how that didn't show up. Added it. Looks good overall, though.

backend/authentication/views.py
markrawls: We should also keep the type as `email` on the frontend for the best UX, we should just always treat data from the frontend as untrusted by default

PR #101, opened 2020-07-17
AUTH: add view to delete user account

### LINKED ISSUE #99: AUTH: add view to delete user account
This is needed for the BofA flow.

laks12metlife:
Closes #99 

Should this be a class-based view instead??  I also tried this out using the DRF Generic DestroyAPIView class and that also worked.  This was how I did it using class-based views:

backend/authentication/urls.py
`path("delete-account/<str:portunus_uuid>/", views.DeleteUserView.as_view(), name="delete_account")`

backend/authentication/views.py
```
class DeleteUserView(DestroyAPIView):
    lookup_field = "portunus_uuid"
    queryset = User.objects.all()
```


cviveiro:
Looking at both options, it looks like there's nothing too out of the ordinary in your function-based view, so I suspect that the class based view will be better because it's more standard, more use of boilerplate, and might handle stuff we're not thinking of handling/work better with outside libraries too.

laks12metlife:
OK makes sense, Thanks!  Changed to class-based view.

backend/authentication/urls.py
```diff
@@ -18,4 +18,9 @@
     path("password-reset/complete/", views.reset_password, name="password_reset_complete"),
     path("change-email/", views.change_email, name="change_email"),
     path("send-new-user-email", views.send_new_user_email, name="send_new_user_email"),
+    path(
+        "delete-account/<str:portunus_uuid>/",
+        views.DeleteUserView.as_view(),
+        name="delete_account",
+    ),
```
Ian-MacLeod, line 25: I think we can convert the existing `RetrieveUserView` to a `RetrieveDestroyAPIView` instead of adding another path here.

laks12metlife: Great point, updated.

PR #120, opened 2020-07-20
FRONTEND: show login settings at root url

### LINKED ISSUE #95: Move account settings page to '/'
The current '/' was meant as a placeholder, and it makes more sense that the account settings live there.

https://github.com/MetLifeLegalPlans/portunus/pull/93#issuecomment-657743424

kborer:
closes #95


frontend/pages/login-settings.js
```diff
@@ -1,3 +0,0 @@
-import LoginSettings from '@@/components/LoginSettings';
```
cviveiro, line 1: This'll need a corresponding change in members

frontend/components/Page.jsx
```diff
@@ -26,10 +26,12 @@ const Page = ({ children, className: externClassName, ...props }) => {
 Page.propTypes = {
   children: PropTypes.node.isRequired,
   className: PropTypes.string,
+  showBack: PropTypes.bool,
 };
 
 Page.defaultProps = {
   className: '',
+  showBack: false,
```
cviveiro, line 34: Other components use this to show a back button. Either set the default to True or change those components.

I think I'd prefer the default -> True

kborer: agreed, updated

PR #122, opened 2020-07-21
AUTH: restrict login attempts after 5 consecutive failures

### LINKED ISSUE #104: H2: lock account after 5 failed authentication attempts
The user should not be able to log in until they reset their password.
When a user's account gets locked, send an email alert with a password reset link.

### report
Screenshot from 2020-07-17 18-37-12

related to https://github.com/MetLifeLegalPlans/willing-app/issues/2138

Ian-MacLeod:
Closes #104 

Restricting just based on account for now.


kborer:
The code looks good but, unfortunately, we know from last time that the reviewers will not accept a lock with a cool off period. Here are the previous PRs:

soft lock that was rejected: https://github.com/MetLifeLegalPlans/willing-app/pull/1968
hard lock with password reset: https://github.com/MetLifeLegalPlans/willing-app/issues/2138

Ian-MacLeod:
This is updated - borrowed heavily from the willing approach.

The task enqueue function might be good to add to zygoat or willing-zg.

backend/shared/utils/tasks.py
```diff
@@ -0,0 +1,26 @@
+from importlib import import_module
```
cviveiro, line 1: Feels like this should be in willing-zg eventually since it keeps getting pulled in.

Ian-MacLeod:
Made a small update here - now calling login_user whenever we change the password, which is necessary because we are blacklisting the existing tokens.

kborer:
lgtm

PR #125, opened 2020-07-21
ADMIN: Add user-specific login settings page for admin access

### LINKED ISSUE #82: ADMIN: Add user-specific login settings page for admin access
Based on the design, an admin should be able to edit a specific user's login credentials.
Looks like we'll need a separate view that handles getting login info by uuid, only accessible by admins.

image

marco-wong-1:
closes #82 

Couldn't find the design for resetting password for specified users so here is what it looks like:

adminreset-password

After hitting the button: 

resetsuccess

Also, I wasn't sure if I needed to create a new file AdminSettingsRow.js. The only difference is the SettingsRow.js routeTo prop is required and the text on the link is "Change" instead of "Request Reset". Possibly didn't need to create a whole new file for this.


backend/authentication/views.py
```diff
@@ -163,6 +163,17 @@ def get_current_user_settings(request):
...
 
+@api_view(["GET"])
+@permission_classes([IsAdminUser])
+def get_user_settings_with_id(request, portunus_uuid):
+    try:
+        user = User.objects.get(portunus_uuid=portunus_uuid)
+        return make_response(data={"email": user.email})
+    except User.DoesNotExist:
+        return make_response()
+    return make_response()
```
cviveiro, line 174: This is not needed. You have returns for all of the different branches.

frontend/components/AdminResetPassword.js
```diff
@@ -0,0 +1,76 @@
...
+  const [email, setEmail] = useState('');
+  const [success, setSuccess] = useState(false);
+
+  useEffect(() => {
+    async function getUser() {
+      try {
+        const response = await getUserSettingsWithId(uuid);
+        setEmail(response.data.email);
+      } catch (e) {
+        console.log(e);
```
cviveiro, line 20: rm

frontend/components/AdminResetPassword.js
```diff
@@ -0,0 +1,76 @@
...
+        )}
+        <Button onClick={handleSubmit} variant="contained" color="primary" noMinWidth size="large">
+          Request Reset
+        </Button>
+        <Spacer v={8} />
+      </div>
+    );
+  };
+
+  const successMessage = () => {
```
cviveiro, line 62: Use `components/Success` for this

frontend/components/UserLoginSettings.js
```diff
@@ -0,0 +1,87 @@
+import { useEffect, useState } from 'react';
```
cviveiro, line 1: This feels so similar to `LoginSettings.js`, I'm sure there are ways to refactor to reuse that and make them look similar... It would just be changing text and then changing what the button does (i.e. what endpoint is hit) right? I don't like having two components that fulfill almost the same functionality with no shared code between. I'm happy to walk through with you if you'd like (not tomorrow though >.<)

backend/authentication/views.py
marco-wong-1: you mean the last return? We'd still want to account for if the user doesn't exist right?

cviveiro: I suspect you've noted this by now, but we should be guarded against the user not existing by the permission class decorator

marco-wong-1:
@cviveiro Is this okay? for the ResetPassword component I am not sure if its still too repetitive.

PR #126, opened 2020-07-22
Restrict login update attempts

### LINKED ISSUE #110: M5: logout user after 5 failed authentication attempts
If a user fails to authenticate 5 times while changing email or password, log them out.

### Report
Screenshot from 2020-07-17 19-07-27

Ian-MacLeod:
Closes #110 

To be merged after #122. Only the last commit is part of this PR.

Kind of awkward adding this to the user model, but putting it in the session like we did for willing would be vulnerable to replay attacks (unless we want to switch to db-backed sessions).


kborer:
I don't think it is awkward -- this is an authentication app after all. Needs rebase.

frontend/components/ChangeEmail.js
```diff
@@ -52,6 +52,11 @@ const ChangeEmail = () => {
 
   const handleError = error => {
     if (error.response && error.response.data) {
+      if (error.response.data.error === AUTH_CHANGE_LOCKOUT) {
+        window.location = '/';
```
cviveiro, line 56: This doesn't seem to show any error for the user explaining why they were logged out. Do we want to add in something showing that? Not in this PR.

Ian-MacLeod: I think something like that could be nice - perhaps using a query parameter and notistack.

PR #130, opened 2020-07-23
FRONTEND: Remove custom server

markrawls:
This preps us for ECS deployment. Just about everything was added through a simple `zg update`.


frontend/package.json
```diff
@@ -7,7 +7,9 @@
     "build": "next build",
     "start": "npm run build && NODE_ENV=production node next-server.js",
```
Ian-MacLeod, line 8: We don't want this to reference the old next-server file, right?

markrawls: Good catch. Updated!

PR #133, opened 2020-07-24
ADMIN: add reset password email endpoint

### LINKED ISSUE #132: add reset password email endpoint
for https://github.com/MetLifeLegalPlans/members/issues/394

this is for admins who want to request an email be sent for another user, so the endpoint should expect to receive a public_id

This is intended to replace sending an admin user to Portunus to reset a user's password.

marco-wong-1:
closes #132


backend/authentication/views.py
```diff
@@ -104,6 +104,23 @@ def request_password_reset(request):
...
+def admin_request_password_reset(request):
+    portunus_uuid = request.data.get("portunus_uuid")
+    if not portunus_uuid:
+        return make_response(False)
+
+    try:
+        user = User.objects.get(portunus_uuid=portunus_uuid)
+    except User.DoesNotExist:
+        # Send back success even if the account DNE to avoid leaking user uuids.
+        return make_response()
```
Ian-MacLeod, line 118: Hmm, I don't think we care about leaking user uuids in an admin view - would probably be better UX to return a failure response here.

marco-wong-1: I just copied the reset-password, but I could change it

hpetrow:
yay. thanks.

PR #139, opened 2020-07-28
API: only send csrf token when it is present

Ian-MacLeod:


Ian-MacLeod:
CSRF is important on some unauthenticated endpoints, too. (login/register for example)

PR #141, opened 2020-07-28
BACKEND: print uuids for users set to staff in management command

### LINKED ISSUE #140: ADMIN: Show UUID when toggle_staff_status is run
We currently don't have an easy way to find someone else's UUID to enable their admin panel access. Show UUID when `toggle_staff_status` is run for an email, then we can use this UUID to set admin status on members. Eventually, we want this to be a one step process. For now, this will help. Need this today

hpetrow:
closes #140


backend/backend/management/commands/set_staff.py
```diff
@@ -30,6 +30,10 @@ def handle(self, *args, **options):
         user_values = [convert(v.strip()) for v in options.get("users")]
         where = {"{}__in".format(options.get("field")): user_values}
         users = User.objects.filter(**where)
+        print(
+            f"Setting staff to {'False' if options.get('unset') else 'True'} for users with uuid(s):",
+        )
         for u in users:
             u.is_staff = not options.get("unset")
+            print(u.portunus_uuid)
```
ambikads, line 38: Can we add a string "UUID" before printing `UUID` so its easier to search/use in the next step?

hpetrow: before each uuid? i thought unformatted might be easier to plug into the members command, since it takes a list separate by space?

Issue #128, opened 2020-07-22
EMAIL: style password reset email


hpetrow:
[Invision](https://projects.invisionapp.com/d/main?origin=v7#/console/19755595/425319686/preview?scrollOffset=741)

PR #147, opened 2020-07-29
Reset password email template styling

metlife-eliles:
#128

### DIFF
--- a/backend/authentication/templates/email/reset_password.html
+++ b/backend/authentication/templates/email/reset_password.html
@@ -4,11 +4,13 @@
   alt="MetLife Logo"
 />
 <br /><br />
-<span style="font-size: medium;">
+<span style="font-size: medium; font-family: sans-serif;">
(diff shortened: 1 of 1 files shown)


backend/authentication/templates/email/reset_password.html
```diff
@@ -1,9 +1,21 @@
...
+  width="200"
+  src="https://www.metlife.com/content/dam/metlifecom/us/icons-header/MetLife.png"
+  alt="MetLife Logo"
+/>
+<br /><br />
+<span style="font-size: medium; font-family: sans-serif;">
 {% if user.name %}
 Hello {{ user.name }},
 <br><br>
-{% endif %}
```
cviveiro, line 4: Don't think this should have been taken out?

metlife-eliles: oh oops you are right, must have fat fingered it away

PR #148, opened 2020-07-29
AUTH: secure user view

Ian-MacLeod:
Not sure how I missed this before..


kborer:
lgtm. Are there any other areas were we should add testing at some point?

Ian-MacLeod:
> lgtm. Are there any other areas were we should add testing at some point?

Looking through the rest of portunus now, will add more if needed.

PR #161, opened 2020-08-01
DEV: use different port for portunus in local development

Ian-MacLeod:
This will make it easier to use both portunus and members locally (probably should have done it earlier). There is also no need to expose the db port in dev.


cviveiro:
Fantastic, thanks!

PR #164, opened 2020-08-06
IE not supported

marco-wong-1:
closes https://github.com/MetLifeLegalPlans/members/issues/547


marco-wong-1:
Haven't tested this yet.

frontend/pages/_document.js
```diff
@@ -7,9 +7,66 @@ export default class MyDocument extends Document {
...
+          <link
+            rel="stylesheet"
+            href="https://cdnjs.cloudflare.com/ajax/libs/outdated-browser/1.1.5/outdatedbrowser.min.css"
+            integrity="sha256-KNfTksp/+PcmJJ0owdo8yBLi/SVMQrH/PNPm25nR/pI="
+            crossorigin="anonymous"
+          />
           <link rel="stylesheet" href="/averta.css" />
         </Head>
         <body>
+          <div id="outdated">
```
cviveiro, line 19: Question: should this be a separate component? Benefit of this could be that we could put it in willing-zg and pull it into other apps, and also it would clean this file up.

marco-wong-1: Yeah I literally just copied and pasted what mark said here: https://github.com/MetLifeLegalPlans/members/issues/547#issuecomment-669610956
I am looking into next.js further to see how I could clean this up though.

marco-wong-1:
I couldn't get the _document stuff to work. Not sure if this is a good way of implementing it. But it should be working. If I change the string to "Mozilla" instead of "Trident" the Modal shows up for me. I'll assume it works on IE.

@markrawls feel free to edit anything
image

frontend/pages/login.js
```diff
@@ -1,20 +1,32 @@
...
   );
 };
 
 Login.public = true;
 
+Login.propTypes = {
+  ua: PropTypes.string.isRequired,
+};
+
+Login.getInitialProps = async ctx => {
```
cviveiro, line 27: I would have thought that this makes more sense to do in the `Login` functional component, rather than as something you can optionally pass in and figuring out the default props dynamically this way? But maybe this was something you did to be able to test?

marco-wong-1: I moved the <BadBrowserModal /> to the functional component, but I think I need to keep the getInitialProps in the pages cause I think thats where next js gets the user-agent

frontend/components/BadBrowserModal.js
```diff
@@ -0,0 +1,44 @@
...
+
+import Typography from '@wui/basics/typography';
+import Modal from '@wui/basics/modal';
+import Button from '@wui/input/button';
+import Spacer from '@wui/layout/spacer';
+
+const BadBrowserModal = ({ userAgentString }) => {
+  const [modalOpen, setModalOpen] = useState(false);
+  useEffect(() => {
+    setModalOpen(userAgentString.match('Trident'));
```
Ian-MacLeod, line 13: we probably want to check for MSIE here, too

frontend/components/BadBrowserModal.js
```diff
@@ -0,0 +1,44 @@
...
+  }, []);
+
+  return (
+    <Modal
+      open={Boolean(modalOpen)}
```
Ian-MacLeod, line 18: modalOpen should already be a boolean, no need to cast.

marco-wong-1:
Yeah, I was having trouble implementing the script blocks in the _document.js and the next.js documentation didn't help that much. I figured this way I could also style it using wui. But when I was trying to do it in _document.js it kept giving me outdatedbrowser not defined and a bunch of other errors. If you could help me get around that, I'll try to do it that way again.

Ian-MacLeod:
>Yeah, I was having trouble implementing the script blocks in the _document.js and the next.js documentation didn't help that much. I figured this way I could also style it using wui. But when I was trying to do it in _document.js it kept giving me outdatedbrowser not defined and a bunch of other errors. If you could help me get around that, I'll try to do it that way again.

If you could push a branch to your remote with where you were at, I'll give it a look.

marco-wong-1:
> > Yeah, I was having trouble implementing the script blocks in the _document.js and the next.js documentation didn't help that much. I figured this way I could also style it using wui. But when I was trying to do it in _document.js it kept giving me outdatedbrowser not defined and a bunch of other errors. If you could help me get around that, I'll try to do it that way again.
> 
> If you could push a branch to your remote with where you were at, I'll give it a look.

I didn't get very far. I kept getting "ReferenceError: addLoadEvent is not defined" and if I define it outside of the return I'd get outDatedBrowser not defined. I googled a bunch and read the documentation on next.js and eventually figured doing it another way would be easier. 

Here is my branch: https://github.com/marco-wong-1/portunus/tree/badbrowser

Ian-MacLeod:
Got around to taking a look at this - I was able to fix the addLoadEvent error and a couple others related to SSR on your branch, which I will push in a new commit. My solution is quite hacky, though, I'd like @markrawls to take a look and potentially suggest a better way.

I had some issues testing this locally - when I try to load Portunus locally in IE11, there is the expected error related to MobX not supporting browsers without Proxy objects, but it seems to prevent any kind of browser paint at all from occurring. Loading `dev` or `prod` in IE11 works as expected, so I'm not sure what to make of this.

Ian-MacLeod:
So it seems that IE fails spectacularly on next running in development mode, but I can get the same behavior as deployed environments if I run it in production mode locally. The updated `badbrowser` branch does work but this one does not - it seems that the Modal component does not work on IE.

markrawls:
NextJS does static optimization in production code, so things that can be (at least initially) static HTML will be. In development mode it doesn't even bother with that, it builds the pages for the first time when you go to them. Which is why you probably see a slight lag before page transitions during development. SSG only happens at build time, which would allow *something* to get to IE

marco-wong-1:
#167 should closes this

PR #167, opened 2020-08-12
Add browser compatibility banner

Ian-MacLeod:
Closes https://github.com/MetLifeLegalPlans/members/issues/547
Supersedes https://github.com/MetLifeLegalPlans/portunus/pull/164

Tested locally with IE11 and this is working. Not very happy with the approach though, would appreciate a look from @markrawls and any suggestions on how to do this better.

### DIFF
--- a/frontend/pages/_document.js
+++ b/frontend/pages/_document.js
@@ -12,7 +12,7 @@ export default class MyDocument extends Document {
             rel="stylesheet"
             href="https://cdnjs.cloudflare.com/ajax/libs/outdated-browser/1.1.5/outdatedbrowser.min.css"
             integrity="sha256-KNfTksp/+PcmJJ0owdo8yBLi/SVMQrH/PNPm25nR/pI="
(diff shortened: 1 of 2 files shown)


Ian-MacLeod:
Thanks, I knew there was a better way. Tested locally and it's working.

frontend/pages/_document.js
```diff
@@ -8,10 +8,39 @@ export default class MyDocument extends Document {
...
+              </a>
+            </p>
+          </div>
+
+          <script
+            src="https://cdnjs.cloudflare.com/ajax/libs/outdated-browser/1.1.5/outdatedbrowser.min.js"
+            integrity="sha256-yV0saZESxHBqfSfNncH044y3GHbsxLZJbQQmuxrXv90="
+            crossOrigin="anonymous"
+          />
+
```
markrawls, line 40: Can you put this after `<NextScript />` so that most users don't have to pause execution to load this before our actual app?

PR #172, opened 2020-08-14
FRONTEND: update styles on the browser compatibility banner

Ian-MacLeod:
We could probably use a better long term solution for this, but this updates the banner styles to match what product wants for it.

### DIFF
--- a/frontend/components/BrowserCompatibilityBanner.js
+++ b/frontend/components/BrowserCompatibilityBanner.js
@@ -1,0 +1,38 @@
+export default () => (
+  <div id="outdated">
+    <div
+      style={{
+        position: 'fixed',
+        top: 0,
+        right: 0,
+        bottom: 0,
+        left: 0,
+        'background-color': 'rgba(255, 255, 255, 0.4)',
+      }}
+    />
+    <div
+      style={{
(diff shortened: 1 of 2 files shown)


frontend/components/BrowserCompatibilityBanner.js
```diff
@@ -0,0 +1,38 @@
...
+        bottom: 0,
+        left: 0,
+      }}
+    >
+      <h6>Your Browser Is Not Supported</h6>
+      <p style={{ 'text-transform': 'none', 'font-size': '14px' }}>
+        Please try another browser to improve your experience and security.
+        <a id="btnUpdateBrowser" style={{ display: 'none' }}>
+          {' '}
+          Learn More{' '}
```
ambikads, line 28: Curious, why the `Learn More` is still here?. I didn't see this in your local screen shot.

Ian-MacLeod: We're using a small js library to show this banner - we wanted to hide this, but it expects this element to be in the DOM (or it errors out and fails). We might want to swap this out for a different method eventually, but this was the easiest way to get it working for now.

PR #176, opened 2020-08-31
FRONTEND: default to register instead of login

kborer:


kborer:
test failure looks unrelated

PR #177, opened 2020-08-31
FRONTEND: remove password confirmation field

kborer:
product team changed their minds again


cviveiro:
I think there are also confirm password textboxes when you reset your password on that page. Should probably be refactored to use the AuthBase.

kborer:
They only care about removing it from registration, so no harm in leaving it in the change password form.

PR #178, opened 2020-08-31
BACKEND: allow login via registration form

kborer:


backend/authentication/views.py
```diff
@@ -36,15 +36,20 @@
...
 
         if not serializer.is_valid():
             first_errors = {k: v[0] for k, v in serializer.errors.items()}
-            return make_response(False, first_errors)
+
+            if backup_serializer_class:
+                serializer = backup_serializer_class(data=data, context={"request": request})
+
+            if not serializer.is_valid():
+                return make_response(False, first_errors)
```
markrawls, line 52: I think the better form of this function is taking `*args` and iterating through each `arg` until one is successful, or returning the errors from the first one if none are successful.

kborer: I've added a second commit to allow an arbitrary number of serializers. If it looks good, I'll squash the commits. If not, I'll drop the latest commit.

cviveiro: I liked the explicitness of having a specified backup class, but I do like being able to take arbitrary serializers. I'd lightly argue for taking in an arbitrary number of backup serializers but keeping them as named/specified backups. I'm not a terribly big fan of having the order passed in matter in a way that you need to read through the code to see.

Then again, this is also something that can be helped by having a docstring. I still think that the multiple backups idea is clearer but docstring would be nice too.

kborer:
Okay, I added a comment and squashed the commits.

PR #185, opened 2020-09-09
Implement CAPTCHA

### LINKED ISSUE #168: Implement CAPTCHA
One of the findings from Synopsys ethical hack tests done for BoA.

Screenshot from 2020-08-13 11-52-25

Screenshot from 2020-08-13 11-51-29

laks12metlife:
Closes #168 

Uses Google's recaptcha v2. 

Before deploying to Prod, the set of API keys need to be changed to the prod set @ambikads.

I don't particularly love the way I'm checking in the backend view if we are registering an user, and therefore need to verify the captcha key.  I was also considering splitting up login and register to 2 separate views for that reason.  Any thoughts would be appreciated!

Couple of misc linting changes, unrelated to issue, that pre-commit hooks updated.

captcha


frontend/components/AuthBase.js
```diff
@@ -11,18 +12,21 @@ import Typography from '@wui/basics/typography';
 
 import TermsCheckbox from '@@/components/TermsCheckbox';
 import { useGlobalContext, useInputFieldState } from '@@/hooks';
+import { GOOGLE_CAPTCHA_SITE_KEY } from '@@/constants';
```
cviveiro, line 15: I think it makes sense, since we need the key in both the frontend and backend, to put it in the backend settings file, and then fetch that for the frontend from the backend envs. That way, we're guaranteed to be working off of the same key and we only have to change it in one place. You can see how it works in members [in this code](https://github.com/MetLifeLegalPlans/members/tree/develop/frontend/utils/env)

Edit: oh looks like this is already existing in portunus! cool, that shouldn't be too bad to add. similar folder structure in portunus

cviveiro (commented):
Awesome!

backend/authentication/views.py
```diff
@@ -46,6 +47,11 @@ def view(request):
             first_errors = {k: v[0] for k, v in serializer.errors.items()}
             return make_response(False, first_errors)
 
+        if request.environ.get("PATH_INFO") == "/api/auth/register/":
+            is_valid_captcha = check_captcha_key(data.get("captchaKey"))
```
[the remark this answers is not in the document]
laks12metlife: Great point.  A captcha key will only be generated on the frontend when the user successfully completely the challenge. So here we are guaranteed to have either a non-null captcha key (on register) or an empty string captcha key (on login).

frontend/components/AuthBase.js
laks12metlife: Actually there are 2 separate keys.  We are using a SITE KEY on the frontend to request the captcha key and a SECRET KEY on the backend to verify the captcha key.  So would it still be best to include both keys in the settings file or just leave them in the respective constants file?

backend/authentication/utils.py
```diff
@@ -177,3 +175,18 @@ def check_change_email_token(token_str, user):
...
+        response = requests.post(
+            url=GOOGLE_CAPTCHA_URL,
+            data={"secret": GOOGLE_CAPTCHA_SECRET_KEY, "response": key},
+        )
+
+        if response.json()["success"]:
+            return True
+        return False
+
+    except requests.exceptions.RequestException:
```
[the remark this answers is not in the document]
laks12metlife: I was just covering my bases here as a catch all to any exception the `requests` module might raise while verifying the captcha key, like a `HTTPError, TimeOutError, ConnectionError` etc.  I imagine they would be pretty rare though.

backend/authentication/tests/views/test_views.py
```diff
@@ -86,7 +86,9 @@ def test_register_existing_email(self, authenticate_and_test):
         data = {**USER_DATA, "password": INVALID_PASSWORD}
         authenticate_and_test("authentication:register", data, success=False)
 
-    def test_redirect(self, authenticate_and_test):
+    @patch("authentication.views.check_captcha_key")
```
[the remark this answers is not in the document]
laks12metlife: OHHH how clever!  I didn't realize this nifty feature.  I set `return_value = True` in the constructor of the `MagicMock`.  That made sense to me, but isn't needed for the tests to pass.  The default return_value is a `Mock` object.  I'm not sure what's best practice here, or if it really matters.

backend/authentication/tests/views/test_views.py
```diff
@@ -167,10 +187,7 @@ def test_create_with_no_existing_user(self):
         new_email = "newuser@test.com"
         request = factory.post(
             reverse("authentication:create_user"),
-            {
-                "email": new_email,
-                "password": VALID_PASSWORD,
-            },
+            {"email": new_email, "password": VALID_PASSWORD},
```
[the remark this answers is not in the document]
laks12metlife: Hmmm just upgrading black isn't telling the pre-commit hooks to use the upgraded version (20.8b1).  It works if I update the pre-commit-config.yaml file to use the latest version.  I also tried pruning my containers using 	`docker system prune -a --volumes` but that didn't work.  Any steps I could be missing?

frontend/components/AuthBase.js
cviveiro: Ahhhh missed that. Separate files on each side seems fine as you have it now!

backend/authentication/tests/views/test_views.py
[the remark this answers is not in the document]
cviveiro: ah, upgrade `flake8-black`, that was what got my pre-commit hooks to install a new version and work i think

backend/authentication/views.py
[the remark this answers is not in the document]
Ian-MacLeod: I think it would be best to add this to the validation in `RegistrationSerializer`. That way it only affects registrations and we can also require the key's existence if that is what we want.

I'm definitely open to splitting these into multiple views if it becomes necessary, but I think it is nicer to put this in the serializer with the rest of the validation logic in any case.

Ian-MacLeod (changes requested):

laks12metlife:
@Ian-MacLeod Good idea. I moved the captcha validation to the RegistrationSerializer, let me know if you'd do anything differently.  `captchaKey = serializers.CharField()` sets `allow_blank` to `False` by default, so if the captcha key is an empty string, the serializer validation will fail.  I created a new set of test inputs `REGISTER_USER_DATA` that includes a test captcha key to include anytime we are `authenticate_and_test`ing the register view.

backend/authentication/tests/views/test_views.py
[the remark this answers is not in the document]
laks12metlife: Argh I can't get it to work without editing the pre-commit file.  `rev: 20.8b1` has already been updated in the pre-commit-config.yaml files in zygoat and members, so is it ok to do so here as well?  If not, we can try to figure it out during out meeting today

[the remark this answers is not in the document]
hpetrow: yeah, the pre-commit file has to have the new version for the hooks to pass. i'll create a PR to update zygoat.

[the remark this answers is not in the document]
hpetrow: zygoat updated. this should be okay now when rebased on latest develop.

Ian-MacLeod (commented):
Looks good to me. Could be nice to test that registration fails with a wrong/missing captcha key

backend/authentication/tests/views/test_views.py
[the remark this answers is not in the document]
laks12metlife: Thanks Heather!!

laks12metlife:
Thanks for the feedback!! Added 2 tests for empty captcha and missing captcha.

There's one thing I realized I need to fix.  After a captcha key attempts to be validated once it is invalid.  So in the case where a new user registers but doesn't provide a sufficient password, completes the captcha, and hits register, that captcha key will no longer be valid.  But the UI doesn't prompt the user to re-do the captcha after they enter a good password.

backend/authentication/constants.py
```diff
@@ -1 +1,3 @@
 MAX_AUTH_CHANGE_FAILURES = 5
+GOOGLE_CAPTCHA_URL = "https://www.google.com/recaptcha/api/siteverify"
+GOOGLE_CAPTCHA_SECRET_KEY = "6Ldof8kZAAAAAB69zPnBf0og7OZyiRzlzmApOhSX"
```
ambikads, line 3: Shouldn't this be a `prod_required_env`?

laks12metlife: Not quite sure that I did this correctly... let me know if I didn't.

cviveiro: What Ambika means is that this should be in the django settings file as a `prod_required_env` and not here. So, when you need to use it, you import the value from settings.

frontend/constants.js
```diff
@@ -3,6 +3,8 @@ export const FACEBOOK_APP_ID = '275311629877038';
 export const GOOGLE_CLIENT_ID =
   '685797762223-edandm45c2023v4bvcqanggdkhgb3ims.apps.googleusercontent.com';
 
+export const GOOGLE_CAPTCHA_SITE_KEY = '6Ldof8kZAAAAAKdQfBfgJQ8Ogbr-esIgLj9K8tYH';
```
cviveiro, line 6: Is this key also something that will also need to get changed when we deploy?

laks12metlife: Yes, that would also need to be changed.  I believe Ambika generated the prod Secret and Site key pair.

cviveiro: Hmm. Then it looks like this should also be in the settings file on the backend where both can be updated together just for prod, and then we fetch it from there on the frontend. So we'll need to put it there and then get it through https://github.com/MetLifeLegalPlans/portunus/blob/develop/frontend/utils/env.js. I'm not too familiar with the process but can help look into it.

laks12metlife:
Still a WIP, no need to review yet.  Shirley is helping me with something :)

laks12metlife:
Moved the captcha api keys to the backend settings file so they use the correct api keys from AWS while running in production.  Added a loading state on the register page to show until the captcha api site key gets set. 

One issue I'm noticing is that in Chromium sometimes, the recaptcha box won't load on first render. When the user starts inputting info or the page gets reloaded, it appears.  On subsequent visits to the register page, this problem doesn't occur.  I haven't experienced this in Firefox.  There's no mention in the docs that the google script needs to be loaded or anything.  It looks like react-google-recaptcha should load the script by itself.  Any ideas on why this might be happening??

no_captchabox
box_appears

cviveiro:
Ahhh I don't know enough to be helpful, @markrawls would you know what's going on here?

markrawls:
That... Is strange. I'm looking into it. If I had to guess it's something to do with SSR?? But that wouldn't explain why it doesn't happen in Firefox.

laks12metlife:
So miraculously after a rebase, I'm not seeing the issue (where the captcha box wouldn't appear on initial render) in Chrome anymore.  Would really appreciate if somebody could also do a sanity check in Chrome.

Made a small fix to the `handleError` method of `AuthBase.js` to only reset the recaptcha ref if `showCaptcha` is true.  Other than that, everything is pretty much the same.

frontend/components/AuthBase.js
```diff
@@ -97,11 +115,19 @@ const AuthBase = ({
...
       next,
+      captchaKey,
     })
       .then(handleSuccess)
       .catch(handleError);
   };
 
+  const onLoading = () => {
+    return <div>Loading...</div>;
+  };
```
Ian-MacLeod, line 126: This function seems unnecessary to me. We can just return this value on line 129 instead.

frontend/components/AuthBase.js
```diff
@@ -42,6 +49,12 @@ const AuthBase = ({
     }
   }, []);
 
+  useEffect(() => {
+    if (showCaptcha && !captchaSiteKey) {
+      env.captchaSiteKey.then(value => setCaptchaSiteKey(value));
+    }
+  }, []);
```
Ian-MacLeod, line 56: Might be worth setting up a mobx store like the `ConfigStore` from members in portunus. Short of that, we can simplify this a little bit:
```
if (showCaptcha) {
  env.captchaSiteKey.then(setCaptchaSiteKey);
}
```

frontend/components/AuthBase.js
```diff
@@ -77,7 +89,13 @@ const AuthBase = ({
   };
 
   const handleError = error => {
-    if (error.response && error.response.data) {
+    if (showCaptcha && !captchaKey) {
+      setInputErrors({ non_field_errors: 'Please complete the captcha to continue.' });
```
Ian-MacLeod, line 93: This seems like a check that should happen in `validateForm` - if the user somehow circumvents that and submits a request without the captcha then the response from the backend should contain the error.

frontend/components/AuthBase.js
```diff
@@ -97,11 +115,19 @@ const AuthBase = ({
...
       .catch(handleError);
   };
 
+  const onLoading = () => {
+    return <div>Loading...</div>;
+  };
+
+  if (showCaptcha && !captchaSiteKey) {
+    return onLoading();
+  }
```
Ian-MacLeod, line 130: This isn't a critical issue, but it could be nice to render the rest of the registration form and just show loading for the captcha input itself while we wait for the captcha key.

backend/authentication/tests/views/test_views.py
```diff
@@ -70,27 +78,40 @@ class TestRegister:
...
-            **USER_DATA,
+            **REGISTER_USER_DATA,
             "email": "abc@abc",
         }
         authenticate_and_test("authentication:register", data, success=False)
 
+    def test_register_with_bad_captcha_key(self, authenticate_and_test):
+        data = {
+            **REGISTER_USER_DATA,
+            "captchaKey": "",
```
Ian-MacLeod, line 96: It would be nice to include a test where the captcha key is included, not blank, but wrong.

backend/authentication/tests/views/test_views.py
```diff
@@ -21,12 +21,19 @@
...
 USER_DATA = {
     "email": USER_EMAIL,
     "password": VALID_PASSWORD,
 }
 
+REGISTER_USER_DATA = {
+    "email": USER_EMAIL,
+    "password": VALID_PASSWORD,
+    "captchaKey": TEST_CAPTCHA_KEY,
+}
```
Ian-MacLeod, line 35: I'd like to see this re-use the USER_DATA fields:
```
REGISTER_USER_DATA = {
    **USER_DATA,
    "captchaKey": TEST_CAPTCHA_KEY,
}
```

Ian-MacLeod (changes requested):

ambikads:
Once code review is done, please don't merge this. Want to make sure Product is up to speed on this.

frontend/components/AuthBase.js
laks12metlife: Thank you Ian!! Sorry for taking so long to get to this.  Updated!  For now, I'll take the shortcut suggestion, but I'll open a separate issue to setup the mobx store for environment vars and assign it to myself, if that's ok with you.  If you'd rather that be in this PR let me know!!

frontend/components/AuthBase.js
laks12metlife: Yup, I agree.  Moved this check to `validateForm`. But it just checks for the presence of a captchaKey, it doesn't submit a request to the backend.  This would be consistent with all the other checks in the `validateForm` function.

frontend/components/AuthBase.js
laks12metlife: Ah, good idea.  Updated!

backend/authentication/tests/views/test_views.py
laks12metlife: Updated!  Previously, I had `patch`ed the function that validates the captcha key for the whole `TestRegister` class, all tests would succeed as long as the captcha key was non-empty.  So now I moved the patch decorator to only the tests that fail without it it: `test_redirect`, `test_redirect_wildcard`, and `test_ignores_bad_redirect_param`.

cviveiro (approved):
lgtm

@laks12metlife should probably get another test since it's been out so long, even though very little has happened to portunus during that time. edit: also rebase!

@ambikads let me know when you've synced up with product and want to get this in.

laks12metlife:
@cviveiro @ambikads tested and rebased

PR #187, opened 2020-09-09
FRONTEND: always use next parameter when present

Ian-MacLeod:


kborer:
lgtm, though not codebuild

kborer (approved):

PR #189, opened 2020-09-09
CI/CD: Bypass ratelimiting

markrawls:
Combined with some updated environment configuration on the codebuild side, we're now using our own "custom" image that authenticates to Docker Hub with a paid account so we no longer get ratelimited. Will be making a similar one for Members.

Note: This is not actually a custom image, it's the exact same as AWS's `ubuntu:4.0` image. They won't let you pull the images directly but they do [provide the source files](https://github.com/aws/aws-codebuild-docker-images) you need to build it yourself.

TL;DR Everything is gonna keep working the same it always has, merge this immediately


Ian-MacLeod (approved):
Lgtm

PR #191, opened 2020-09-11
BACKEND: blacklist tokens on logout even when authorization header no…

Ian-MacLeod:
…t present

Currently, we log users out by redirecting them to the log out path - this means that the authorization header is not present in the request. To make sure that we blacklist all the tokens for the user this acquires the user from the refresh token in the session cookie.


cviveiro (approved):
LGTM

For my own benefit, when would the authorization header not be included? Was this causing bugs or are you making this change because it's just the right thing to do?

PR #193, opened 2020-09-14
LOGGING: add logging for various auth events

Ian-MacLeod:
The new logging module provides two helpers for adding logging: the `log_event` function, which can be called directly to log an event, and the `log_view_outcome` decorator, which can be added to a view to log basic information about the response.

Currently, this logs the status code, timestamp, ip address, and user email for the decorated views.

If this looks good I will add it to the willing-zg django app and we can use it in other zygoat repos, too.

I'm using the `log_event` function for password reset requests, because it currently does not send error responses when the email is not found to avoid leaking user data.


cviveiro (approved):
I like! Making it easy and standard to log is always good in my view :)

PR #195, opened 2020-09-16
BACKEND: swap out zygoat-managed python modules for django equivalents

### LINKED ISSUE #194: Update Portunus to use zygoat-django and django-willing-zg
The django apps should be ready to swap in for the zygoat-managed python code.

Steps
1. Run zg update
2. Change all usages of zygoat-managed python modules to import from `zygoat_django` or `willing_zg` instead.
3. Delete the existing zygoat-managed python modules

AFAIK that should be it, @markrawls Let me know if I missed anything here.

JCog:
closes #194


backend/backend/settings.py
```diff
@@ -1,12 +1,12 @@
...
 import json
-import boto3
+import os
 import environ
+import boto3
+import sentry_sdk
+from datetime import timedelta
+
+from sentry_sdk.integrations.django import DjangoIntegration
+from zygoat_django.settings import *  # noqa
```
Ian-MacLeod, line 9: it looks like we're overwriting a bunch of stuff from django-zygoat in this settings file, I think we want to just let this import grab what it can, which means this settings file will look a lot more like the previous `settings/__init__.py` file, without the stuff from `zygoat_settings.py`

Ian-MacLeod (changes requested):

Ian-MacLeod (approved):
Lgtm

PR #197, opened 2020-09-17
AUTHENTICATION: add list users view

### LINKED ISSUE #181: Add API endpoint to fetch batch emails for a list of IDs
This is useful for the admin tools in Members and will be useful for the other sites we link to Portunus.

hpetrow:
closes #181


backend/authentication/tests/views/test_views.py
```diff
@@ -168,44 +166,59 @@ def test_logout_blacklists_user_tokens(self, client, authenticate_and_test):
...
+    def test_list_all_users(self):
+        self.check_list()
 
-        self.assertEqual(response.status_code, 400)
-        self.assertEqual(
-            json.loads(response.content)["portunus_uuid"], staff_user.portunus_uuid
-        )
-        self.assertEqual(json.loads(response.content)["user_exists"], True)
+    def test_list_users_by_uuid(self):
+        self.check_list(True)
```
cviveiro, line 216: This to me is a perfect use of `parameterized`, because I also got caught by the fact that the `check_list` function really looked like a test rather than a utility function and I was about to comment on that.
```
from parameterized import parameterized
...
@parameterized.expand(
    [
        (False,),
        (True,),
    ]
)
def test_list_users(self, from_uuid):
    [ all of the code in `check_list` ]
```

backend/authentication/urls.py
```diff
@@ -9,7 +9,7 @@
     path("register/", views.register, name="register"),
     path("login/", views.login, name="login"),
     path("users/settings/", views.get_current_user_settings, name="get_current_user"),
-    path("users/", views.CreateUserView.as_view(), name="create_user"),
```
cviveiro, line 12: We'll need to be extra careful about deploying this with members changes and making sure the versions in the different apps all work together, right?

backend/authentication/tests/views/test_views.py
hpetrow: cool, thanks! updated.

backend/authentication/urls.py
hpetrow: definitely. this shouldn't affect the create users portunus call from the members app, because it should still be a POST to /users, but will definitely want to test.

backend/authentication/views.py
```diff
@@ -201,9 +200,17 @@ def post(self, request, *args, **kwargs):
...
-    serializer_class = CreateUserSerializer
+    serializer_class = UserSerializer
+
+    def get_queryset(self):
+        uuids = self.request.GET.getlist("portunus_uuids")
+
+        if uuids:
+            return User.objects.filter(portunus_uuid__in=uuids)
+
+        return User.objects.all()
```
Ian-MacLeod, line 213: Are we concerned about responding with all user info unpaginated when uuids are not provided? Maybe responding with no user info in this case is better?

Ian-MacLeod (commented):
Lgtm, just one question

backend/authentication/views.py
hpetrow: sure, wouldn't hurt. updated.

cviveiro (approved):
LGTM, don't know what's going on with camel case but  _shrug_

backend/authentication/tests/views/test_views.py
```diff
@@ -223,7 +232,7 @@ def check_get(self, status_code):
         self.assertEqual(response.status_code, status_code)
 
         if status_code == 200:
-            self.assertDictEqual(response.json(), UserSerializer(self.user).data)
+            self.assertDictEqual(response.data, UserSerializer(self.user).data)
```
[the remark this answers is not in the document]
Ian-MacLeod: So, it seems that before #195 we were not using the camel case renderer/parser. I made a PR to address this (#204 ) and after that is merged, this should pass using response.json() (which I think is preferable)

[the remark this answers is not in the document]
hpetrow: cool, got it. we still need to use `response.data` here, because the camel case renderer is still being used in the `RetrieveDeleteUserView`, but changed back to `response.json()` for the `ListCreateUsersView` tests. should be good now. thanks!

Ian-MacLeod (approved):
Lgtm

PR #202, opened 2020-09-21
BACKEND: restore DATABASES setting

JCog:


Ian-MacLeod (approved):
Lgtm

backend/backend/settings.py
```diff
@@ -86,6 +86,8 @@ def get_secret(secret_arn):
 # Database
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
+DATABASES = {"default": db_config}
+
```
Ian-MacLeod, line 90: Actually, it looks like this should be below line 104, right?

Ian-MacLeod (changes requested):

backend/backend/settings.py
JCog: oh, you know you're right, that'd probably be better

Ian-MacLeod (approved):

PR #203, opened 2020-09-21
FRONTEND: cleanup filenames and constants, add jest setup

swang192:


frontend/jest.config.js
```diff
@@ -0,0 +1,189 @@
+// For a detailed explanation regarding each configuration property, visit:
```
Ian-MacLeod, line 1: Do we want to keep all the commented out settings in this file?

frontend/jest.setup.js
```diff
@@ -0,0 +1,40 @@
...
+  useGlobalContext.mockReturnValue(window.store);
+
+  window.mockedRouter = {
+    route: '',
+    pathname: '',
+    query: '',
+    asPath: '',
+    push: jest.fn(() => new Promise(r => r())),
+  };
+  nextRouter.useRouter = () => window.mockedRouter;
```
Ian-MacLeod, line 39: I'm sure that there must be a better way to make these objects available during tests than attaching them to the window, but I don't know what that is offhand. I suppose I'm ok with this for now, but we should look for a nicer solution.

Ian-MacLeod (commented):
Looks good

frontend/components/Nav.js
```diff
@@ -19,7 +19,7 @@ import Typography from '@wui/basics/typography';
 
 import { useToggledState, useGlobalContext } from '@@/hooks';
 import NavMenu from '@@/components/NavMenu';
-import { SUPPORT_PHONE_NUMBER } from '@@/constants';
+import { SUPPORT_PHONE_NUMBER } from '@@/utils/constants';
```
markrawls, line 22: The '@@/' is no longer necessary in the new version of `NextJS` with the `jsconfig.json` file that we have

frontend/jest.config.js
markrawls: In my opinion we should, but I am not inherently correct

markrawls (changes requested):

markrawls (approved):

PR #204, opened 2020-09-22
API: accommodate camel case renderer and parser for API calls

Ian-MacLeod:
Portunus was apparently not using the camel case parsers and renderers before #195 - the drf settings were being overridden in `settings/__init__.py`. This updates the API usages to use camel case for request parameters and response parsing. Most of these requests would actually be fine unchanged because they would not be affected by the parser, but I figure we should use it as intended.

I'm updating the create/delete user view to keep using the old parser/renderer for now because members currently expects snake case in the response.

There are a couple bugs that this fixes, too - the unknown error response on the change email/change password pages was not working because the state key being set was not used. Also, the complete email change page had a mismatch in the name of the prop being passed/received.

### DIFF
--- a/backend/authentication/views.py
+++ b/backend/authentication/views.py
@@ -5,6 +5,7 @@ from calendar import timegm
 from django.views.decorators.http import require_POST
 from django.core.exceptions import ValidationError
 from rest_framework.generics import CreateAPIView, RetrieveDestroyAPIView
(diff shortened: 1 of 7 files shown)


kborer (approved):

JCog (approved):
lgtm

PR #208, opened 2020-09-22
FRONTEND: change sentry env variables to camel case

JCog:
As an extension to #204 this also changes the sentry environment variables to camel case instead of snake case to be parsed correctly


laks12metlife:
Would https://github.com/MetLifeLegalPlans/portunus/blob/80d8530e22dc74937f889fd732f973239d511981/frontend/pages/_app.js#L29 also need to be updated in this PR?

Ian-MacLeod:
> Would
> 
> https://github.com/MetLifeLegalPlans/portunus/blob/80d8530e22dc74937f889fd732f973239d511981/frontend/pages/_app.js#L29
> 
> also need to be updated in this PR?

Yep, looks like it! Just updating these would work, but the best solution would probably be best to use the keys constant from `env.js` in this `Promise.all`.

JCog:
ok, this should be fixed now

Ian-MacLeod (approved):
Lgtm

Ian-MacLeod (changes requested):
Oh, wait, I've misled you. This should actually just be the sentry dsn and env so using the keys constant isn't good. Sorry!

JCog:
> Oh, wait, I've misled you. This should actually just be the sentry dsn and env so using the keys constant isn't good (works for now, but could easily break if the keys constant changes). Sorry!

Ah, makes sense! And no problem, updated.

Ian-MacLeod (approved):

PR #215, opened 2020-09-29
Nav Metlife Logo leads to home queryparam

### LINKED ISSUE #166: LOGIN SETTINGS: Take the user to members services page when logo is clicked
Clicking the logo in Login settings page from Members App should take the user to Services page.

https://www.notion.so/Update-logo-link-to-return-to-Category-Page-from-Login-settings-page-b18bbf65bfcf4504a191da1c66ffdb54

marco-wong-1:
closes #166 after implementing query param on members.
Users shouldn't be able to get to portunus without coming from members (or future apps) but in case that happens the logo would link to home of portunus. (not sure if I need to account for this)


Ian-MacLeod (changes requested):
This looks good, but we should validate the content of the home query parameter to make sure that it is a legalplans.com subdomain.

marco-wong-1:
so I'm not too familiar with how we structure the domains, but I just used a regex to make sure it contains one legalplans.com in there. @Ian-MacLeod if you want to take another look 😃

frontend/components/Nav.js
```diff
@@ -49,10 +48,19 @@ const useStyles = makeStyles(theme => ({
...
   const menuText = 'My Account';
-  const { query } = useRouter();
+  const router = useRouter();
+  const { home } = router.query;
+
+  useEffect(() => {
+    const re = /^.*(legalplans\.com){1}.*$/;
+    if (re.test(home)) {
+      setHomeLink(home);
+    }
```
Ian-MacLeod, line 62: I think that using the URL interface to grab the host name would be a bit nicer. Also, we should also not allow things like `notlegalplans.com`. Something like this would work:

```
const hostName = new URL(home).hostname;
if (hostName.endsWith('.legalplans.com')) {
```

Ian-MacLeod (changes requested):

marco-wong-1:
Forgot to checkout to a new branch so closing this

PR #211, opened 2020-09-24
RC20200924

ambikads:

PR #216, opened 2020-09-29
FRONTEND: use port 3000 for local and deployed frontend

Ian-MacLeod:


ambikads:
lgtm as long as this is tested locally. nit: port needs to be updated on 2 other files that I see from the previous [commit](https://github.com/MetLifeLegalPlans/portunus/commit/8294ac5a31e8e5333e07e66070c7553ac397e224) - `e2e.yml` (should add this in CI once it runs actual tests) and `frontend/README.md`

And for reference, this is to fix the 502s we saw while deploying https://github.com/MetLifeLegalPlans/portunus/pull/211

Ian-MacLeod:
Working locally. As for the two other files - I think port 3001 is correct for portunus e2e once we start using it. The reverse proxy still listens on port 3001 locally. I was thinking the same thing for the frontend readme, but the whole file is not very useful, actually, because we don't tend to run the frontend outside of it's docker container - it might be best to just remove it and add something more relevant to the root readme. I suppose 3000 is slightly more accurate in the readme as is, so I'll revert it for now.

ambikads (approved):

PR #223, opened 2020-10-09
FRONTEND: make zendesk script the direct child of nextjs Head

Ian-MacLeod:
Script tags should be direct descendants of Head, see:
https://nextjs.org/docs/api-reference/next/head


swang192 (approved):

kborer:
Does this fix some bug or improve behavior? When `ChatWidget` was previously being included in `Head`, was it not just turning into a script tag anyway? Asking for a friend.

ambikads:
I noticed that when the login page loads or refreshed for the first time, the intermediate UI screen kind of looks weird- the logo appears on the top and banner is wider etc..think the screen that flashes before login app is from members and hence the weird refresh look.

ambikads:
@Ian-MacLeod this PR did not fix the issue we discussed.

PR #224, opened 2020-10-11
TESTING: add cypress e2e tests

### LINKED ISSUE #39: Create e2e tests

swang192:
Close #39 

Will rebase when the zendesk fix is in


cviveiro (commented):
the tests themselves look fine to me, wish we could do tests around the email functionality but i guess these on the frontend and the unit tests on the backend should have us mostly covered.

looks like most of this should be going into zygoat though so that other apps get it. cypress as a dev dependency sounds fine to me.

swang192:
I added just one test suite so far for the happy path, mainly to avoid repeatedly hitting the same components and increasing the test run time. Was planning to test edge cases and error states using Jest.

Was the email functionality you were thinking of the email/password reset pages with the `uuid` in the pathnames? I could try to add a separate test suite for admin features.

PR #225, opened 2020-10-12
BACKEND: add email search API endpoint

### LINKED ISSUE #182: Add email search API that returns portunus uids
This is something that we couldn't do with Members but would like to be able to in the future. We need the puids to link the Portunus user (which includes email) with the Members user (which doesn't include email). We'll need to be thoughtful about the search, normalize parts of it or make it take characters like `.` or caps well. We'll want to pass in the search string and return the puids of the matching users.

I think @hpetrow has specific ideas on how the search should work, if she wants to add them here.

kborer:
#182


backend/authentication/tests/views/test_views.py
```diff
@@ -217,6 +217,42 @@ def test_unauthenticated_user(self):
...
+                True,
+                False,
+            ),
+            (
+                True,
+                True,
+            ),
+        ]
+    )
+    def test_search_email(self, from_email, partial_email):
```
cviveiro, line 241: i find these vars to be confusingly named. Would rather `partial_email` -> something like `use_partial_email`. Not sure about a better name for `from_email`. `use_search_arg` but that's a bit bleh.

backend/authentication/views.py
```diff
@@ -243,6 +246,23 @@ def create(self, request, *args, **kwargs):
             return make_response(False, data)
 
 
+class SearchUsersView(ListAPIView):
```
hpetrow, line 249: Not sure about this, but couldn't this be part of  `ListCreateUsersView`, since I think `ListCreateAPIView` handles list api view?

backend/authentication/tests/views/test_views.py
kborer: agreed, updated

backend/authentication/views.py
kborer: I didn't do that because it uses a different queryset

backend/authentication/views.py
```diff
@@ -243,6 +246,23 @@ def create(self, request, *args, **kwargs):
...
+    permission_classes = [IsAdminUser]
+    serializer_class = UserSerializer
+    renderer_classes = [JSONRenderer]
+    filter_backends = [
+        filters.SearchFilter,
+    ]
+    search_fields = SEARCH_FIELDS
+
+    def get_queryset(self):
+        search = self.request.GET.get("search")
```
Ian-MacLeod, line 259: I think `request.query_params` is slightly preferable to `request.GET`

Ian-MacLeod (changes requested):
Looks good. We may want to consider placing a limit on the number of results.

kborer:
Thanks, I've implemented your suggestions and also added a minimum length for the search string.

cviveiro (approved):
LGTM, not sure if Heather or Ian have more to say

PR #227, opened 2020-10-12
FRONTEND: look for snake cased non_field_errors in login/register api…

### LINKED ISSUE #226: No error message when login API request returns an error response code
I can see this on Staging and not on Prod, so possibly related to https://github.com/MetLifeLegalPlans/portunus/pull/222

Screenshot from 2020-10-12 14-51-10

Ian-MacLeod:
… responses

The login and register views are implemented as vanilla django views
to ensure that CSRF protection is applied as recommended in the DRF
docs. This means that the camel case to snake case custom renderers
are not used to build the response and we need to check for
`non_field_errors` and not `nonFieldErrors`.

Closes #226

### DIFF
--- a/frontend/components/AuthBase.js
+++ b/frontend/components/AuthBase.js
@@ -80,7 +80,7 @@ const AuthBase = ({
     if (error.response && error.response.data) {
       setInputErrors(error.response.data);
     } else {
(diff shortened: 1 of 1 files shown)


kborer (approved):

cviveiro (approved):
Good catch. LGTM.

PR #228, opened 2020-10-13
password length is 8 chars

### LINKED ISSUE #198: Password minimum length should be set to 8 characters on the backend
Currently at 7. Taking this out from #169 so that it can be separate and a good beginner issue, mostly for learning how development works and the org of the repo.

pbirl:
closes #198 
**Change Password** - Change password screen shot
**password 7 chars on register**- password lenght 7
**Password length 8 chars** - password length 8
**Reset password for existing account with 7 chars-**reset password

### DIFF
--- a/backend/authentication/tests/views/test_views.py
+++ b/backend/authentication/tests/views/test_views.py
@@ -65,7 +65,7 @@ class TestRegister:
             "test_password",  # No number
             "837493938481",  # No letter
             "password1",  # Too common
-            "uam38m",  # Too short
+            "uam38m2",  # Too short
         ],
     )
     def test_register_with_bad_password(self, authenticate_and_test, password):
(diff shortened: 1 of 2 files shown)


laks12metlife (approved):
Looks good to me!

cviveiro:
Changes lgtm and thanks for changing the tests to make sure!!

You should try and follow [these guidelines](https://chris.beams.io/posts/git-commit/#seven-rules) when writing your commit message. It will make this commit fit in with the rest of our commit history.

So for this, usually we like to prepend our commits with the general area where you're doing things, "BACKEND" or "AUTH" would work here, since you're working in the backend on some authentication stuff. So change your commit name to something more like this: "AUTH: change minimum password length to 8 characters". And then automatically your PR is also named that :)

For doing that, you'll want to run `git commit --amend`. This'll allow you to rename the commit you have in this PR. I think it should open up nano for you by default to do things, you'll want to rename the message in nano and then ctrl-x and ctrl-y to close and save, which'll save your changes (i think). If you have questions, lmk!

If you hate nano (it's not great), you can configure your gh to use something else: https://www.oreilly.com/library/view/gitlab-cookbook/9781783986842/apas07.html

Issue #183, opened 2020-09-08
    BACKEND:add stricter permissions on delete user API endpoint


cviveiro:
This should be relatively easy. We'll want stricter permissions than the normal API endpoints, probably superuser instead of staff.

cviveiro:
for the record, we already have this endpoint but the permissions aren't strict enough, so this is now making it so that only superusers can delete a user, rather than all staff.

PR #232, opened 2020-10-19
BACKEND: Add delete user API endpoint

pbirl:
#183


backend/authentication/factories.py
```diff
@@ -15,3 +15,8 @@ class Meta:
 
 class StaffUserFactory(UserFactory):
     is_staff = True
+
+
+class SuperuserFactory (UserFactory):
```
Ian-MacLeod, line 20: Code style change: remove the space between `SuperuserFactory` and the `(`

Ian-MacLeod (changes requested):
Looks good, one linting error - we should get pre-commit set up in your environment so the linter will automatically run, but in the meantime I left a comment that should fix the error.

There is a separate commit here from your minimum password length PR - rebasing this branch off of the latest version of `develop` should fix that.

backend/authentication/tests/views/test_views.py
```diff
@@ -256,7 +256,13 @@ def test_same_user(self):
...
     def test_staff_user(self):
         staff_user = StaffUserFactory()
-        self.check_user(staff_user)
+        self.client.force_authenticate(staff_user)
+        self.check_get(200)
+        self.check_delete(403)
+
+    def test_superuser(self):
+        superuser = SuperuserFactory()
+        self.check_user(superuser)
```
cviveiro, line 265: this isn't related to the changes this PR makes, but i wish `check_user` was named a bit more descriptively, so that it's clear what it's doing. maybe more like `check_user_actions`?

backend/authentication/factories.py
```diff
@@ -15,3 +15,8 @@ class Meta:
 
 class StaffUserFactory(UserFactory):
     is_staff = True
+
+
+class SuperuserFactory(UserFactory):
```
cviveiro, line 20: Nit: I'd prefer this `SuperUser` instead of `Superuser`, and same throughout, to match more with the `StaffUser`

backend/shared/permissions.py
```diff
@@ -4,3 +4,8 @@
 class IsSameUserOrAdmin(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
         return obj == request.user or request.user.is_staff
+
+
+class IsSameUserOrSuperuser(permissions.BasePermission):
```
cviveiro, line 9: Same comment here re: `SuperUser`

cviveiro:
Also, we're trying to start being better about adding in descriptions for what a PR does. I can add it in when merging, but it's good practice to go further than just saying what issue is closed by it and describing what it does. You can do something like this for this one:
```
This tightens the permissions on the delete user endpoint to be superuser rather than all staff.
```
Simple but will let non-engineers know exactly what changed.

backend/authentication/factories.py
Ian-MacLeod: I suggested this capitalization while we were pairing to be more consistent with the `is_superuser` field name (I think superuser is usually one word, too?) - I don't feel too strongly either way, though.

cviveiro: Meh then don't worry about changing it I also don't feel strongly

Ian-MacLeod:
Looks good, but the commit history is a little funky now - also needs to be brought up to date with `develop`. Let me know if you need help with this.

cviveiro (approved):

cviveiro (approved):

PR #241, opened 2020-12-04
SETTINGS: add support phone number and email address and use in email templates

### LINKED ISSUE #155: SETTINGS: add support phone number and use in email templates
After adding the env var, we can use the `support_phone_number` var in reset password email.

pbirl:
closes #155 
Support phone number and email address

### DIFF
--- a/frontend/components/Nav.js
+++ b/frontend/components/Nav.js
@@ -11,6 +11,7 @@ import Divider from '@material-ui/core/Divider';
 import PhoneIcon from '@material-ui/icons/Phone';
 import ExpandLessIcon from '@material-ui/icons/ExpandLess';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import * as Sentry from '@sentry/react';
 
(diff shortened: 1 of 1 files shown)


backend/authentication/templates/email/reset_password.html
```diff
@@ -15,7 +15,7 @@
 <br><br>
 <a href="{{ reset_url }}"><button style="background-color: #3C59BA; color: #ffffff; border-radius: 3px;">Reset Password</button></a>
 <br><br>
-If you need more assistance, please feel free to reach us at +1 800-821-6400 or clientservice@legalplans.com
+  If you need more assistance, please feel free to reach us at {{support_phone_number}} or {{support_email_address}}
```
cviveiro, line 18: remove the spaces put at the beginning of this line.

also on this line, i think the standard for us and for django templating is to add spaces between the `{{` and `}}` and the var inside, so `{{ support_phone_number }}` and `{{ support_email_address }}`.

PR #242, opened 2020-12-07
Send invalid URL error to sentry

marco-wong-1:
If the home parameter is an invalid URL. we should send an error to sentry as @Ian-MacLeod suggested


Ian-MacLeod (approved):
Looks good

PR #243, opened 2020-12-08
BACKEND: Prep for ECS migration

markrawls:
# DO NOT MERGE THIS YET, IT WILL BREAK MOST THINGS

## It is ready for review though, so do that please

Sets a good production Dockerfile, updates the dependencies, implements Celery (without need of a separate container in prod), and moves our singular task over to a Celery `shared_task`.


frontend/zg_utils/google-analytics.js
```diff
@@ -1,6 +1,6 @@
 import ReactGA from 'react-ga';
 import Router from 'next/router';
-import { GOOGLE_ANALYTICS_ID } from 'utils/constants';
+import GOOGLE_ANALYTICS_ID from 'utils/constants/google';
```
Ian-MacLeod, line 3: Is this right? I think it would still need to be destructured.

docker-compose.yml
```diff
@@ -1,12 +1,17 @@
 services:
   backend:
-    build:
+    build: &id001
```
Ian-MacLeod, line 3: Could use some more descriptive names for these anchors.

Ian-MacLeod (changes requested):
Looks good in general.

This includes a zg update, right? I think it's nicest if we can put the results of that in their own commit for clarity.

frontend/zg_utils/google-analytics.js
markrawls: Yep, that is correct

docker-compose.yml
markrawls: This is what Prettier generated for them in the pre-commit hook, didn't catch those at first

backend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7
+FROM python:latest
```
ambikads, line 1: I am not so sure if we want this, at least not until we get to practice "build once, deploy many" approach.

backend/Dockerfile.local
```diff
@@ -7,4 +7,4 @@ ADD requirements.txt /code/
 ADD requirements.dev.txt /code/
 RUN pip install -r requirements.dev.txt
 
-CMD ./manage.py wait_for_db && ./manage.py migrate && ./manage.py createcachetable && ./manage.py runserver 0.0.0.0:3001
+CMD ./manage.py wait_for_db && ./manage.py migrate && ./manage.py createcachetable && uvicorn --host 0.0.0.0 --port 3001 --reload backend.asgi:application
```
ambikads, line 10: naive question, does celery not work for local development and testing?

markrawls: It does! We're running it in a separate instance of the backend container though, so it reloads every time we make a change. The production daemon runs in the same container but does not provide reloading.

backend/Dockerfile
markrawls: Wouldn't that be what we're moving to with the backend on ECS?

ambikads: It wouldn't. We will need to change the way we deploy to production. Once the images are built for testing, the same image should be deployed to production. Right now we build every time.

markrawls: Ah, noted. I'll pin it at 3.9 for now.

markrawls:
> This includes a zg update, right? I think it's nicest if we can put the results of that in their own commit for clarity.

That was only dependency updates, and splitting those out manually is a touch difficult. For the time being I'm gonna opt to leave em in one commit.

Ian-MacLeod:
> That was only dependency updates, and splitting those out manually is a touch difficult. For the time being I'm gonna opt to leave em in one commit.

It's OK to leave it in here since they are already mixed, but in the future I think a separate commit makes sense. In particular, it helps grokking what is included when looking at the commits in a RC.

frontend/public/browserCompatibility.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
Ian-MacLeod, line 1: Did this sneak in here somehow? It's not needed, right?

Ian-MacLeod (changes requested):

frontend/public/browserCompatibility.js
markrawls: I put that in there because the linter wanted it moved to ES6 standards and this file doesn't get transpiled.

Ian-MacLeod (approved):
This looks good to me

markrawls:
Excellent, I'm going to get this merged shortly and work on moving over some lower environments.

ambikads:
fine with me too. When ready, lets merge to develop and fix the CI/CD and deploy to dev before proceeding with other lower environments(qa and test). Please remember to treat staging similar to production as we have some sponsors using testing for various tests.

PR #249, opened 2021-01-07
SETTINGS: add support phone number and email address and use in email templates

### LINKED ISSUE #155: SETTINGS: add support phone number and use in email templates

pbirl:
#155


cviveiro (approved):
lgtm!

PR #250, opened 2021-01-07
 BACKEND: Password strength on backend should be restricted to a reasonably good score

### LINKED ISSUE #199: Password strength on backend should be restricted to a reasonably good score
With #174 we're thinking we'll start restricting passwords to need to be good enough before they get accepted. [zxcvbn](https://github.com/dropbox/zxcvbn) is a great way to do this that doesn't just rely on length and including specific characters, but rather takes a holistic score based on multiple metrics. We're thinking we'll restrict to a score of 3 on the frontend currently (I think). We should do the same on the backend.

This will mean creating our own password validator that will only allow passwords with a score higher than that we define. For an example of rolling out our own, we had to do it earlier with portunus [adding an alphanumeric restriction](https://github.com/MetLifeLegalPlans/portunus/pull/13/files#diff-8d989e96dcebafe306fe370788384403). Also see [django's documentation on writing validators](https://docs.djangoproject.com/en/3.1/ref/validators/#writing-validators)

pbirl:
#199

### DIFF
--- a/backend/authentication/password_validators.py
+++ b/backend/authentication/password_validators.py
@@ -2,10 +2,12 @@ import re
 
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext as _
+from zxcvbn import zxcvbn
 
 
 has_alphabetic = re.compile(r"[a-zA-Z]")
 has_numeric = re.compile(r"[\d]")
+MIN_SCORE = 3
 
 
 class AlphaNumericPasswordValidator:
@@ -24,3 +26,17 @@ class AlphaNumericPasswordValidator:
             "Your password must contain at least 1 numeric character and 1 alphabetic "
             "character"
         )
+
+
+class PasswordStrengthValidator:
+    help_string = """Your password needs to be strong.
+                  Try making it longer or adding in numeric and special characters to make it stronger."""
+
+    def validate(self, password, user=None):
+        score_data = zxcvbn(password)
+        score = score_data.get("score")
+        if score < MIN_SCORE:
+            raise ValidationError(self.help_string)
+
+    def get_help_text(self):
+        return _(self.help_string)
--- a/backend/backend/settings.py
+++ b/backend/backend/settings.py
@@ -127,6 +127,9 @@ AUTH_PASSWORD_VALIDATORS = [
     {
         "NAME": "authentication.password_validators.AlphaNumericPasswordValidator",
     },
+    {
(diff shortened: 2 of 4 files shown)


backend/authentication/password_validators.py
```diff
@@ -24,3 +26,20 @@ def get_help_text(self):
             "Your password must contain at least 1 numeric character and 1 alphabetic "
             "character"
         )
+
+
+class PasswordStrengthValidator:
+    def validate(self, password, user=None):
+        help_string = "Your password needs to be strong." \
```
cviveiro, line 33: this is currently only defined within the context of this `validate` function, which means that when you try and use it in `get_help_text` that that won't work. to get that to work, move defining `help_string` to outside of this function, i.e.
```
class PasswordStrengthValidator:
help_string = "Your password......."

def validate(self, ...):
...
```
and then it should all work together!

backend/authentication/password_validators.py
```diff
@@ -24,3 +26,20 @@ def get_help_text(self):
...
+                      "Try making it longer or adding in numeric and special characters to make it stronger."
+        score_result = zxcvbn(password)
+        keys_values = score_result.get('score')
```
cviveiro, line 36: you'll find that i pick on names during code review a lot because i do think good naming (which is hard, it is) really helps code to become more readable.

so, for this, `key_values` doesn't tell me what this is and in fact it is kind of more confusing. this value is the score, so i think you should rename `key_values` to what it is, `score`. similarly, in the line above, `score_result` is pretty good but given that it's a lot of data about what score the password got, a better name might be `score_data`. that'll also clue people into why you have to pull the score out of that dictionary, because it's a lot of data about the score and not just the resulting score.

hope that makes sense!

cviveiro (commented):
awesome! a couple of things

backend/authentication/password_validators.py
pbirl: Oh yeah ...got it i will change that

cviveiro:
this lgtm as is, but there's a frontend change we should make to make sure that this error message (which i don't want to cut down because it's helpful to users) doesn't change the form size. also, we might want to look into https://github.com/MetLifeLegalPlans/portunus/issues/251 before putting this out.

pbirl:
Hey @cviveiro i fixed all the issues. I think it's ready to merge into develop

cviveiro (approved):

PR #252, opened 2021-01-13
BACKEND: do not lock out registration attempts for accounts that do n…

Ian-MacLeod:
…ot exist

When the registration serializer fails validation, a login attempt is still made. This update will prevent the login attempt for accounts that do not exist. Lockouts will still happen when an email for an existing account is submitted (resulting in a real login attempt).

Other solutions to this could be to remove the login-using-register-form behavior, or to make sure that the field level validation is identical on the frontend and backend (which would prevent the registration serializer validation failures).

### DIFF
--- a/backend/authentication/serializers.py
+++ b/backend/authentication/serializers.py
@@ -26,6 +26,8 @@ class RegistrationSerializer(serializers.ModelSerializer):
 
 
 class LoginSerializer(serializers.ModelSerializer):
+    bad_credentials_error = _("Invalid username or password")
+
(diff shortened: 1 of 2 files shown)


cviveiro (approved):
ahhh, so it looks like the key here was to avoid hitting an error with `authenticate` because that's what axes looks for. thanks! i like this change as it is.

i definitely don't like the mix of register and login on the front but that was a product decision and we would have to check with them if we want to change it back.

lgtm!

PR #253, opened 2021-01-13
BACKEND: Cleanup for ECS Migration

markrawls:
Looks like we were encountering some kind of race condition with token blacklisting under serverless, so that needed to be moved inside the `try`/`catch` block. Also there were migrations that weren't generated so I did those as well.

Outside of that though, the move to ECS was super easy!


JCog (approved):
lgtm

PR #258, opened 2021-02-04
BACKEND: Close account enumeration on registration

markrawls:
Just gives a very generic error response instead.


backend/authentication/views.py
```diff
@@ -60,6 +60,12 @@ def view(request):
                 break
 
         if not serializer.is_valid():
+            if not show_errors:
+                return make_response(
+                    False,
+                    {"non_field_errors": "An unkown error has occurred. Please try again."},
```
cviveiro, line 66: "unknown"

markrawls: Whoops, my bad!

PR #261, opened 2021-02-04
PRE_COMMIT: fix pre-commit config

Ian-MacLeod:
pre-commit fails to install prettier when using the latest version
of Node. Advice found here was to upgrade the prettier version and
point to this pre-commit specific repo:
https://github.com/prettier/prettier/issues/9459


cviveiro:
this is a zygoat file, you should just `zg update`. there are some other improvements/updates too to this file that are already reflected in that. https://github.com/MetLifeLegalPlans/zygoat/pull/133. sorry, i was supposed to zg update all the repos and then i got lazy and then i got busy.

Ian-MacLeod:
>this is a zygoat file, you should just zg update. there are some other improvements/updates too to this file that are already reflected in that. MetLifeLegalPlans/zygoat#133. sorry, i was supposed to zg update all the repos and then i got lazy and then i got busy.

Ah, thanks, will do the update instead.

PR #268, opened 2021-02-10
Implement invisible CAPTCHA

### LINKED ISSUE #168: Implement CAPTCHA

laks12metlife:
closes #168 and would be used instead of #185

This uses Google's reCaptcha v2 invisible badge version.  There is no checkbox the user has to click, but a picture challenge will pop up if there's suspicious activity.  

Only implemented on the `/register` page, but since Login and Register are now coupled on that page, user will have to face reCaptcha on both actions.  

** Some notes:
- Sometime the little recaptcha badge on the bottom right corner doesn't show up.  On page load sometimes it flickers and then doesn't appear.  Maybe it has to do with SSR or changing routes?  I'm not really sure yet, I'm still looking into it (any ideas as to what might be going wrong appreciated!).  Whether the badge is there or not doesn't affect the functionality.
- I had to make those edits to `.pre-commit-config.yaml` for my pre-commit to work, but I can get rid of those once #263 gets merged.

image


backend/authentication/serializers.py
```diff
@@ -6,17 +6,26 @@
...
 
-    def validate_password(self, value):
-        validate_password(value)
-        return value
+    # CaptchaKey is marked invalid after one attempted validation.
+    # Raise other errors first because we want to check if LoginUsingRegisterSerializer is valid with the same captchaKey.
+    def validate(self, data):
+        validate_password(data["password"])
+        if User.objects.filter(email=data["email"]).first():
+            raise ValidationError("User with email already exists.")
```
ambikads, line 25: Does this error get displayed to the  user? If so, we don't want to do that. See https://github.com/MetLifeLegalPlans/portunus/pull/258 which is to show a generic error response. Showing email already exists in our system is a security concern.

backend/authentication/serializers.py
```diff
@@ -50,7 +61,8 @@ class LoginUsingRegisterSerializer(LoginSerializer):
     def validate(self, data):
         if User.objects.filter(email=data["email"]).first() is None:
             raise ValidationError(self.bad_credentials_error)
-
+        if not check_captcha_key(data["captchaKey"]):
+            raise ValidationError("The captcha key is not valid.")
```
ambikads, line 65: Do we know how this library handles multiple failed captcha challenges?

laks12metlife: So the frontend will actually continue to prompt the user with picture challenges until it's sure that the user is a human.  So really the frontend will handle multiple failed challenges.  The backend (server-side) validation is to ensure that the key was indeed generated by google and is not a fake key.

